PR #5, opened 2016-04-08
Sl/calib

spencerlyon2:
This implements a new set of calibration types.

We now have

```
immutable ModelCalibration
    flat::FlatCalibration
    grouped::GroupedCalibration
    symbol_table::Dict{Symbol,Tuple{Symbol,Int}}
    symbol_groups::OrderedDict{Symbol,Vector{Symbol}}
end
```

Where `FlatCalibration` and `GroupedCalibration` are light wrappers around `OrderedDict` and `Dict` that allow us to customize getindex and setindex! behavior. 

Semantics have changed a little bit for getindex and setindex! for `ModelCalibration`:

- `getindex` is only implemented on symbols and simply forwards on to the grouped set. To get values for specific variables, use `getindex` on the `flat` field
- `setindex!` is also only implemented where keys are symbols, but allows you to set single variables or groups.

### DIFF
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -7,7 +7,7 @@ using NLsolve
 
 export AbstractModel, AbstractSymbolicModel, AbstractNumericModel, ASM, ANM,
        AbstractDoloFunctor, SymbolicModel, DTCSCCModel, DTMSCCModel,
-       ModelCalibration,
+       FlatCalibration, GroupedCalibration, ModelCalibration,
 
        # functions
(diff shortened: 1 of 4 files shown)


spencerlyon2:
CI failure on julia 0.5 is from a dependency that fails to install.

albop:
This is a very cool API. I tested it on my machine and it works like a charm. Merging it now.

spencerlyon2:
Nice, thanks! Glad you liked it and that it works.

PR #6, opened 2016-04-09
Sl/model tidyup

spencerlyon2:
I do a few things here:

- Add `name`, `filename`, `model_type`, `options`, `distribution` to number model types
- The `options` and `distribution` fields are exact replicas of the corresponding fields in the symbolic model, but all symbols are replaced with numerical values:
```
julia> dump(sm.options)  # symbolic model
Dict{Symbol,Any} len 1
  Approximation: Dict{Symbol,Any} len 3
    orders: Array(Any,(2,))
      1: Int64 10
      2: Int64 50
    b: Array(Any,(2,))
      1: ASCIIString "1+2*sig_z"
      2: ASCIIString "k*1.1"
    a: Array(Any,(2,))
      1: ASCIIString "1-2*sig_z"
      2: ASCIIString "k*0.9"

julia> dump(m.options)  # numeric model
Dict{Symbol,Any} len 1
  Approximation: Dict{Symbol,Any} len 3
    orders: Array(Any,(2,))
      1: Int64 10
      2: Int64 50
    b: Array(Any,(2,))
      1: Float64 1.032
      2: Float64 10.290476119160585
    a: Array(Any,(2,))
      1: Float64 0.968
      2: Float64 8.419480461131387
```
- Added tests for symbolic model constructor
- Added methods `eval_with(mc::ModelCalibration, x)` where `x` is many other things like dicts, arrays, numbers, etc. Calling it on a container type (e.g. Dict) recursively calls `eval_with` until we get down to numbers wherever possible. The numeric version of `sm.options` is obtained by doing `eval_with(calibration, sm.options)`
- Added tests for dtcscc model constructor
- Added tests for dtcscc functions

### DIFF
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -10,7 +10,7 @@ export AbstractModel, AbstractSymbolicModel, AbstractNumericModel, ASM, ANM,
        FlatCalibration, GroupedCalibration, ModelCalibration,
 
        # functions
-       eval_with, evaluate, evaluate!
+       eval_with, evaluate, evaluate!, model_type, name, filename
 
 # set up core types
 abstract AbstractModel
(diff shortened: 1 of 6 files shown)


albop:
Excellent. So you can already import the rbc model, which is really nice.
Also, I hadn't seen the mutating version `evaluate!` until now.
I'll merge that now.

spencerlyon2:
Yep, I should be able to import almost all of the example models!

PR #7, opened 2016-04-12
Sl/linear solve

spencerlyon2:
Very simple linear solver for dtcscc models.

Test it with the rbc model using:

```julia
using Dolo
using YAML: load_file
# replace with path to rbc model in your dolo.py
src = load_file("/Users/sglyon/src/Python/dolo/examples/models/rbc.yaml")
sm = SymbolicModel(src, :dtcscc)
m = DTCSCCModel(sm)
ts = linear_solve(m)
```

Then `ts` is an object of type  `TaylorExpansion`, which has `call` implemented for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ os:
 julia:
   - release
   - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 notifications:
   email: false
 # uncomment the following lines to override the default test script
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -1,13 +1,17 @@
 module Dolo
 
+using Base.Cartesian
+
 using MacroTools
 using DataStructures: OrderedDict
 using YAML: load_file
 using NLsolve
+import ForwardDiff
+using Calculus
 
 export AbstractModel, AbstractSymbolicModel, AbstractNumericModel, ASM, ANM,
        AbstractDoloFunctor, SymbolicModel, DTCSCCModel, DTMSCCModel,
-       FlatCalibration, GroupedCalibration, ModelCalibration,
+       FlatCalibration, GroupedCalibration, ModelCalibration, TaylorExpansion,
 
        # functions
        eval_with, evaluate, evaluate!, model_type, name, filename
@@ -33,6 +37,8 @@ const RECIPES = _symbol_dict(load_file(joinpath(src_path, "recipes.yaml")))
 include("util.jl")
 include("parser.jl")
 include("model_types.jl")
+
+include("numeric/taylor_series.jl")
 
 include("algos/dtcscc.jl")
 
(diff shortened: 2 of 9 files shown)


albop:
One part which looks a bit strange to me is the representation of taylor series. I don't see the need to fill higher order derivatives with matrices.

spencerlyon2:
I'm not sure I follow what you are saying here. 

Are you talking about how for a 1st order `TaylorExpansion` I set `x_2` to `zeros(0, 0, 0)` an d`x_3` to `zeros(0, 0, 0, 0)`?

If so, I think we have a couple alternatives:

- Have different types for each order. The separate types would have only the needed fields.
- Let `x_2` and `x_3` be `Nullable{Array{Float64,3}}` and `Nullable{Array{Float64,4}}` respectively. For 1st order both would be null, for second order only `x_3` would be null, for 3rd order none would be null.

The issue I have with the first alternative is that will cause us to have more types than needed. Not a show stopper, but also not ideal (in my opinion).

The issue I have with the second alternative is that it makes things _slightly_ more complicated as we have to think about packing and unpacking these arrays in `Nullable`s. Not a huge deal, but it does make things a little less straightforward.

I settled on the current implementation because it doesn't come with a memory cost (we just make empty matrices with the correct number of dimensions) and doesn't requires us to deal with multiple types or wrapping/unwrapping in nullable.

I'm open to discussion either way.

Also, I think I'll delete the vectorized version of `Base.call` for now as it isn't clear what convention we want to adopt regarding if a matrix of states should have variables go across rows (implemented here) or down columns (standard almost everywhere else).

spencerlyon2:
Ok I think everything is finished with this PR except for resolving the discussion on what do do with the `x_ss` and `x_sss` arrays on lower order `TaylorExpansion`s

@albop I'll wait to hear your response to the ideas I proposed above.

albop:
@spencerlyon2 : 
So, correct me if I'm wrong, the problem stems from the fact that you cannot get several instances of the parameterized type to have different fields.
I agree that from a coding perspective, different types are less elegant than one parameterized type. But the latter is a shortcut for the former isn't it ?
From a user perspective, there would be no disadvantage at all to have several types (potentially derived by the same one). 
The nullable fields, do not provide much value to me, w.r.t. an empty matrix:  you still see some kind of a dummy field when you look at your object.
I don't feel very strong about multiple types, vs parameterized so I'm not going to argue much more and let you decide. Until we find the perfect type which matches a series of matrices of increasing dimensions...

albop:
Actually, to keep things moving I think I will merge it as it is. When/if we find a perfect solution, we'll revisit.

spencerlyon2:
I'm on the subway on my phone,  so I'll have to elaborate more, but another option would be to use an inner constructor and incomplete initialization. We could then just leave the unused arrays as undefined from Julia's perspective. 

This is my least favorite option of the 4 we have discussed, but it does sort of address your comment about parametric types with different fields.

PR #8, opened 2016-04-12
albop/yaml import

albop:
This adds a simple `yaml_import(url)` which returns a numeric model. `url` can be either a local file, or a remote one.


src/model_import.jl
```diff
@@ -0,0 +1,40 @@
...
+        data = YAML.load(buf)
+    else
+        data = YAML.load_file(url)
+    end
+
+    # fname = basename(url)
+
+    model_type = guess_model_type(data)
+
+    sym_model = SymbolicModel(data, model_type, "filename")
```
spencerlyon2: Can we do better with the filename here? Shouldn't it be at the end of `url` just before `.yaml` or `.yml`?

test_yaml.jl
```diff
@@ -0,0 +1,53 @@
+using YAML
+using Dolo
+
+data = YAML.load_file("/home/pablo/Programming/econforge/dolo/examples/models/rbc.yaml")
```
spencerlyon2: Should we copy some of the example models into this repo and then get this path using `Pkg.dir("Dolo", "examples", "models", "rbc.yaml")`?

Then these test can actually pass on CI

test_yaml.jl
```diff
@@ -0,0 +1,53 @@
...
+
+N = 100
+
+
+S = repmat(s',N)
+X = repmat(x',N)
+E = repmat(e',N)
+P = repmat(p',N)
+
+evaluate(f,S,X,E,S,X,P)
```
spencerlyon2: This almost certainly won't work as I've restricted the implementation of `evaluate` and `evaluate!` to only accept `AbstractVector` so far

albop:
I think we should do it for exactly one model, or at most one per model type.
Otherwise, pulling from the web seems like a good way to go. In the long term, we probably want a model database that lives outside of both dolo.jl and dolo.py

spencerlyon2:
model database is a great idea.

For CI testing that we can import models from the filesystem I agree that one model of each type is perfect

src/model_import.jl
```diff
@@ -0,0 +1,40 @@
+using YAML
```
spencerlyon2: Instead of putting this here, lets just change the line in src/Dolo.jl to be

```julia
using YAML: load, load_file
```

src/model_import.jl
```diff
@@ -0,0 +1,40 @@
...
+using Requests
```
spencerlyon2: Similar to above, let's move this to src/Dolo.jl and have it read `using Requests: get`

spencerlyon2:
We also need to add `Requests` to the REQUIRE file

albop:
Is it a convention that all package imports moving to the top-level file ? It then makes it harder to know which function comes from which package.
Anyway, I don't mind, so I did as you say.

spencerlyon2:
Yeah, it's a pretty strong Julia convention. For a while I didn't like it because as you say it makes it not obvious where functions used in a particular file come from. It's grown on my since then and now I do like being able to open up the top-level file and immediately seeing all imported dependencies and exports

spencerlyon2:
Looks good to me, thanks!

PR #9, opened 2016-04-13
add inline documentation

abhijithch:
I am trying to add inline documentation for the exported APIs, it would help if the authors can review and comment. Thanks.

### DIFF
--- a/src/model_types.jl
+++ b/src/model_types.jl
@@ -127,7 +127,7 @@ end
 
 """
 A dictionary which provides values for parameters and variables.
-Grouped calibrations takes only scalars and vectors as values.
+Grouped calibrations takes only vectors as values.
 ### Example:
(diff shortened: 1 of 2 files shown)


src/model_types.jl
```diff
@@ -107,12 +112,25 @@ end
...
+"""
 FlatCalibration() = FlatCalibration(OrderedDict{Symbol,Float64}())
 
 immutable GroupedCalibration <: Associative{Symbol,Vector{Float64}}
     d::Dict{Symbol,Vector{Float64}}
 end
 
+"""
+A dictionary which provides values for parameters and variables.
+Grouped calibrations takes only scalars and vectors as values.
```
albop: I think it is "vectors as values" instead of " scalars and vectors as values" ?

albop:
Thank you @abhijithch . This is always useful ! Just remember that some of the objects are still moving targets and may evolve a bit...

PR #10, opened 2016-04-13
WIP: first implementation of vfi for dtmscc

albop:
So far the focus is on correctness, so I haven't really looked at performances, tried to make the code especially beautiful or given any thought to user facing options.

Two remarks concerning external libraries:
- interpolations.jl: does not seem to do cubic splines on a space other than `[-1,1]^d` . This will ultimately come, but I have used `splines.jl` in the meantime. Linear interpolations seems to pose some problems to the optimizer.
- Optim.jl has an fminbox algorithm doing constrained maximization which is what we need. It has not been updated yet to use the general `optimize(...)` api. According to this discussion  https://github.com/JuliaOpt/Optim.jl/issues/179, `fminbox` is still work in progress and there isn't (yet) a general constrained optimization framework. I am not 100% sure, but I may have come across a small bug: optimization always fail in the last iteration, that is when the point starting the optimization is nearly optimal.

Also, to make progress, I have quickly amended the recipes to add a `felicity` function and a `rewards` symbol groups. I have also, temporarily, left the markov chain definition to the model.options field.


src/algos/dtmscc.jl
```diff
@@ -0,0 +1,286 @@
...
+    it = 0
+    maxit = get(options, :maxit, 500)
+    tol =   get(options, :tol, 1e-6)
+    err = 10
+
+    dims = vcat(approx.orders)
+    knots = [collect(linspace(approx.a[i], approx.b[i], approx.orders[i])) for i=1:d]
+    knots = tuple(knots...)
+
+    create_interpolant(vv) = [interpolate(knots, copy(reshape(slice(vv,i_mc,:),dims...)), Gridded(Linear()))  for i_mc in 1:n_mc ]
```
spencerlyon2: Here you are using `Gridded(Linear())` because the knots are all originally `linespace` objects, it would be more efficient to use 

```
knots = [linspace(approx.a[i], approx.b[i], approx.orders[i]) for i=1:d]
scale(interpolate(copy(reshape(slice(vv,i_mc,:),dims...)), BSpline(Linear()), OnGrid()), knots...)
```

Not only will this be more efficient, but scaling works on any interpolation types, so you can do things like 

```
scale(interpolate(copy(reshape(slice(vv,i_mc,:),dims...)), BSpline(Cubic(Flat())), OnGrid()), knots...)
```

src/algos/dtmscc.jl
```diff
@@ -0,0 +1,286 @@
...
+
+    diff = 100
+    diff_0 = 1.0
+    diff_2_0 = 1.0
+
+    dims = vcat(approx.orders)
+    knots = [collect(linspace(approx.a[i], approx.b[i], approx.orders[i])) for i=1:d]
+    knots = tuple(knots...)
+
+    # interpolations.jl does not have cubic gridded interpolations yet
```
spencerlyon2: See comment above about scaled Cubic interpolation, which will work here because all knots are subtypes of `Range`

src/algos/dtmscc.jl
```diff
@@ -0,0 +1,286 @@
...
+    # using splnes.jl isntead in the meantime
+     create_interpolant(vv) = [interpolant_cspline(approx.a, approx.b, approx.orders, copy(reshape(slice(vv,i_mc,:),dims...) )) for i_mc in 1:n_mc]
```
spencerlyon2: Instead of doing copy + slice we can just index because indexing copies by default

spencerlyon2:
A few general comments:

- `src/dtcscc_model.jl` was an old file. Let's delete it
- I think you included the example yaml file at both the top level and inside the `examples` directory.

Overall this looks good. I don't have the time to look through it very carefully right now and test it out, but I like how it is looking so far.

I'll comment when I have more time to do a better review

src/algos/dtmscc.jl
albop: Actually, I didn't really want to copy the array, but noticed that some of the functions called below didn't work with slices. Will do what you suggest.

albop:
Some comments from today's progress:
- I have added a simple `time_iteration` method which uses the vectorized variant of the model functions. Again, it is not meant to be especially optimized or user friendly. It doesn't deal with occasionally binding constraints yet.
- this one pulled a few new files with it:
    - `serial.jl` : specialized operations (multiplication and newton optimization), for functions `Nxp->Nxp` when the derivative is sparse with blocks of size p (i.e. each point could be solved separately).
    - `misc.jl` : mlinspace and placeholders for markov chains and approximation spaces
    -  `policies.jl` : objects represent several types of decision rules (simple, indexed by markov state, constant, ...)

Note that the purpose of `serial.jl` can later be fulfilled by using the appropriate type of sparse matrices and the right solvers. Before looking that, I will first port the occasionally constraint solver from dolo in serial.jl though.

As for `policies.jl`, I think we will need this abstraction since  we don't want to deal with completely different api depending on what kind of interpolation we happen to use. I intentionally kept a convention very close to `python/dolo`, but we should discuss how to evolve it. In particular, we could have, instead of `set_values` methods like `regress`, `fit` or `learn`. But we will need to think of the vocabulary very carefully.

PR #11, opened 2016-04-13
WIP: vectorized eval

spencerlyon2:
The goal of this PR is to allow us to do vectorized evaluation of the model functions. The implementation allows a mix between some arguments being vectors (single points for each variable) and other arguments being matrices (multiple points for each variable).

This required some significant changes to the parser/compiler, including:

- swapping all arithmetic operators `+`, `-`, `*` `/`, and `^` for their element wise counterparts: `.+`, `.-`, `.*` `./`, and `.^`. This will have no performance hit on single point evaluation because those methods on scalars are lowered/compiled to the original non-elementwise operator
- We now have functions for unpacking function arguments. This allowed me to `@inline` a very simple function that dispatches on the type of the argument to either extract an element of an input vector (for single point evaluation) or a whole column of the input matrix (for vectorized evaluation).
- Similar point for the output argument to each function. We `@inline`d a simple function to set one element at a time for Vector output and one column at a time for matrix output.
- We now check the sizes of the output array to make sure the size conforms, given the arguments
- We check that all AbstractMatrix arguments have the same number of observations. 
- Removed `@inbounds` from the unpacking statements. we don't check the the sizes are correct, so we can't really guarantee that `@inbounds` is allowed
- For functions with a target, we now check that the order of equations is consistent with the declaration order of the variables
- I cleaned up the generated code a bit to make it more readable. We don't have quite as many `begin...end` blocks.

I also fixed the `print_code` keyword arg when constructing the numeric models and added that as a keyword argument to `yam_import`

The convention I adopted, which I believe is what we settled on for `TaylorExpansion`, is that in a Matrix argument or output, a column would represent all observations of a single variable. If this isn't correct let me know

### DIFF
--- a/src/model_import.jl
+++ b/src/model_import.jl
@@ -11,7 +11,7 @@ function guess_model_type(data)
     end
 end
 
-function yaml_import(url)
+function yaml_import(url; print_code::Bool=false)
 
     if match(r"(http|https):.*", url) != nothing
         res = get(url)
@@ -28,9 +28,9 @@ function yaml_import(url)
     sym_model = SymbolicModel(data, model_type, fname)
 
     if model_type == :dtcscc
-        return DTCSCCModel(sym_model)
+        return DTCSCCModel(sym_model; print_code=print_code)
     elseif model_type == :dtmscc
-        return DTMSCCModel(sym_model)
+        return DTMSCCModel(sym_model; print_code=print_code)
     else
         throw(Exception)
     end
--- a/src/model_types.jl
+++ b/src/model_types.jl
@@ -379,7 +379,7 @@ for (TF, TM, ms) in [(:DTCSCCfunctions, :DTCSCCModel, :(:dtcscc)),
         Base.convert(::Type{SymbolicModel}, m::$(TM)) = m.symbolic
 
         # model type constructor
-        function Base.convert(::Type{$TM}, sm::SymbolicModel)
+        function ($TM)(sm::SymbolicModel; print_code::Bool=false)
             if model_type(sm) != model_type($TM)
                 msg = string("Symbolic model is of type $(model_type(sm)) ",
                              "cannot create model of type $($TM)")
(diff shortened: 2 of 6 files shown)


albop:
Thank you. Thanks to this one we'll be able to make immediate progress on
vectorized algorithms.
I like the inline trick to adapt to the input dimension. Then it
essentially means that the casting rules are implemented by the dotted
operators ?
In the long run we will want to try other vectorization schemes and avoid
creating temp. arrays. Using @devec for each equation may already give
smthg similar to python's numexpr.
I'll still merge it as it is since we don't need further optimizations
right now.

spencerlyon2:
That's right, all the casting magic happens based in Julia's definitions of the dotted operators.

In another PR we should try using `@devec` to see what happens. It might just work out of the box, but it might also have issues.

albop:
As I said above it's not very urgent and we might want to wait to have a
more modular structure so that we can measure costs and benefits
scientifically (with/without on several models for instance).
I was also wondering how a simple loop around the non vectorized fun would
perform. This wouldn't have cache issues and if the compiler inlines the
function it could end up to be very efficient.
Bottom line is that there are several options and no need to rush.

spencerlyon2:
That's definitely another option and I totally agree there are lots of different experiments we can run, but not right now. Having some models/algorithms in place using this functionality will provide a natural environment for doing these benchmarking comparisons down the road.

One last comment: one benefit to having the functions defined on `AbstractMatrix` is that if some day we utilize a library like ArrayFire.jl for GPU computations, we don't have to modify anything.

albop:
Can you do Cuda kernels with array fire ? Does it do some kind of lazy
evaluation ?

spencerlyon2:
You can swap out for 3 different backends: CPU bound, CUDA, OpenCL

I don't know the details behind how computation happens, but @ranjanan might.

I do know that you instantiate an array fire array, which allocates on the GPU. Then the array never comes back to the CPU unless you explicitly ask it to.

albop:
Yes apparently the c++ lib does lazy evaluation so that might work,

PR #12, opened 2016-06-07
Sl/rewrite compiler

spencerlyon2:
Here's a (certainly incomplete) list of the changes I made here

- removed `AbstractDoloFunctor` type
- removed `DTCSCCfunctions` and `DTMSCCfunctions`
- Added type parameters for `ANM`, `ASM`, `DTCSCCModel` and `DTMSCCModel`
- exporting functions for all model functions (e.g. `arbitrage`, `transition`)
- `symbolic` field on all <: ANM is now ASM
- added rbc.yaml to tests dir


albop:
I'll have a look soon. 
@spencerlyon2 : have you run some performance tests to see if removing the functors, did or did not have an impact with julia 0.5 ?

spencerlyon2:
I haven't run any performance tests, but I'm fairly confident that it would break down like this (on either 0.4 or 0.5):

- Time to construct a `DTCSCCModel` or `DTMSCCModel` should be lower as we aren't creating 8 or 9 new types (one for each functor) each time a model is built.
- Time to evaluate the model functions should be exactly the same. Instead of doing `evaluate(functor, args...)` we now do `function_name(model, args...)` -- both are just method calls and should have the same performance characteristics. 

There are a lot of additional benefits this brings:

One of the benefits is that it is much easier to build upon this system and define custom models by hand. Users can simply define their own subtype of `AbstractNumericModel` and then add methods to the functions necessary to use the Dolo functionality they want. E.g. they would only need to implement `arbitrage!` and `transition!` to get a linear solver for dtcscc models. Before they would have had to build a `DTCSCCFunctions` instance that filled in those two fields with callable types and set all the other fields to `nothing` (or some other throw-away value).

Another benefit is we have much less compilation. The `DT_SCCfunction` types had at least 8 type parameters, which required an un-necessary amount of specialization for that type as well as the associated `DT_SCCModel` types, which took a type parameter that fully specified the 8+ function type parameters.

albop:
Oh, I am completely sold on the benefits of it. I was just wondering whether the recent changes in Julia made the need for functors completely disappear, or just mostly disappear.

spencerlyon2:
On 0.4 functors are mostly useful because passing functions as arguments to other functions wasn't well optimized. On 0.5, passing functions is now much faster so functors aren't needed.

We were using them for a totally different thing. I was trying to embed the functions in a type, which just feels wrong in Julia. It makes more sense and is more efficient (on any version of Julia) to add a type parameter to the type and use dispatch to identify a method than embedding the method into the type itself. 

I think the introduction of functors in Dolo.jl was just a result of me over-engineering a bit... I'm happy with this alternative though!

spencerlyon2:
@albop can this be merged. I'm basing all future work on it and don't plan on going back to the old compiler?

albop:
Yes of course, I just did it.

PR #13, opened 2016-07-07
Yaml tags

spencerlyon2:
This does a few things:

- Together with [this commit of YAML.jl](https://github.com/spencerlyon2/YAML.jl/commit/d9cf218f68c42822a39a8a3ad4319d801f8dfafe) we can now handle the tags in the YAML file and build out the objects defined there. I think the fix on the YAML.jl side is kinda hacky and we should probably come up with a better way to allow this functionality. When (if) we do that we will need to revise here.
- I've moved the distribution settings back inside options as in dolo.py. This means it is no longer a field on `DTCSCCModel` and `DTMSCCModel`.
- Added types for the grid and distribution options as well as an Options type. `sm.options` will still be a dict for `sm <: SymbolicModel`. `m.options` is now an instance of `Options` for all the `m <: AbstractNumericModel`we define here.

### DIFF
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -10,6 +10,7 @@ using NLsolve
 using QuantEcon
 using Distributions: MvNormal
 import ForwardDiff
+import YAML
 
 export AbstractModel, AbstractSymbolicModel, AbstractNumericModel, ASM, ANM,
        SymbolicModel, DTCSCCModel, DTMSCCModel,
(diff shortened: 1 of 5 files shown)


albop:
Thank you @spencerlyon2 . The change in YAML.jl makes me very happy.
The Travis failure doesn' t seem related to the work in this branch so that I'll merge it right now.

spencerlyon2:
Thanks for merging. I believe the travis failure is due to one of my testing examples not being updated for moving distributions inside options. I'll try to fix it up soon

PR #14, opened 2016-07-08
Sl/definitions

spencerlyon2:
This is a fairly big set of changes that does the following:

- updates recipes.yaml to match what is currently in dolo.py
- Supports replacing definitions. I have chosen to do the replacements at the time the functions are compiled. If this is the wrong point in the compilation sequence let me know and we can discuss where to move it.
- Update rbc.yaml to match current python version

Things we need to do after this (I'll make issues for each of them, I just want to document them here because they first became known issues while working on this PR):

- learn how to handle the `arbitrage_exp` equation family in dtcscc (and possibly dtmscc -- I haven't checked it if is there)
- Build a `MarkovChain` type so we can import dtmscc models that use the `!MarkovChain` tag.


spencerlyon2:
NOTE: travis failures are due to YAML.jl not supporting custom tags. Tests pass locally for me

spencerlyon2:
I don't think there was anything special with `arbitrage_exp` since they don't allow complementarities. I've enabled them, so once this is merged we can close #15

albop:
Cool thanks. 
Indeed, I think the substitution should be used when the function is compiled so that we can have something like in dolo/compiler/function_compiler.py: 
`make_function(equations, arguments, parameters, targets=None, rhs_only=False, definitions={}, funname='anonymous')`

(where equations is a list of expressions and arguments smthg like: {s=>[(:k,1),(:z,1)], x:=>[(:i,1)]} ). This turns out to be super useful, and having the ability to add definitions there makes it... well... even more useful.

Just a quick question: in your implementation, are you substituting the definitions in each equation symbolically (by changing the ast) ? If so, they might end up repeated many times. In general it *might* be more efficient to just prepend the definitions in the code you generate. We might still want to substitute them if we are to compute higher order derivatives (otherwise we need to deal with complicated chain rules).

spencerlyon2:
I am substituting by operating on the ast. 

What do you mean by "put the definitions in the code I generate"?

PR #19, opened 2016-07-14
ENH: added model file to parse Dynare .mod files

jagman88:
I've tested this on the two .mod files in examples/models/dynare, but we should test it on a few more. I can get on to this next.


albop:
Thanks @jagman88 , I'll have a look right now

albop:
@jagman88 : I have a few comments

- currently the import method doesn't run because it doesn't deal correctly with initval/endval/shocks blocks when they are not there. In this case, you probably want to return empty dictionaries. That way you would do for instance:

jagman88:
I already return empty dictionaries for initval and endval when they don't appear in the .mod file. I can do the same for shocks as well?

jagman88:
Oh! I just realized, this is an old version. Let me update now.

albop:
Ok then I'll wait

spencerlyon2:
@jagman88 can I get you to start a new pull request that targets the `dynare` branch instead of the master branch?

albop:
Looks good to me. Two very minor comments:
- if you match a given block  like in `m = match(r"endval;(.*?)end;", text)`, it is redundant to check `contains(text, "endval")`, you can just do `if m != nothing ...` and gain 0.001 s
- I would tend to close the files, just after the readlines instructions so that I am sure it is eventually closed, even if some disaster appears during the processing

Do you think it will work with the EA_QUEST3 model, that we use to benchmark compilation times ?

albop:
Another one, (for the new PR targeting dynare):  can you move the import function from the test folder to the dolo folder and add it to the files that are included upon import ?
Another easy thing to do is to check the output of the parser against the predicted one in `test/dynare_parser.jl`

jagman88:
@spencerlyon2 can I leave that last request to you? 

I'm currently working through some examples, and doing this check. The problematic example is going to be EQ_QUEST3.mod, which is a bit messy. Will add some fixes before pushing again. 

@spencerlyon2 you can close this PR now, if you like, and move the conversation over to the dynare branch.

PR #20, opened 2016-07-14
ENH: added model file to parse Dynare .mod files

jagman88:
[Moved pull request to the dynare branch]

I've tested this on the two .mod files in examples/models/dynare, but we should test it on a few more. Currently working on testing against EQ_QUEST3.mod (so, so messy...).

### DIFF
--- a/test/modfile_parser.jl
+++ b/test/modfile_parser.jl
@@ -15,52 +15,58 @@ using DataStructures
 function modfile_parser(mod_file_name)
   f = open(mod_file_name)
   lines = readlines(f)  # reads in line by line
+  close(f)
 
-
-  # Remove lines beginning with comments
+  # Remove lines beginning with comments (%, //, /*, \*)
   filter!(x -> ~ismatch(r"^\s*%", x), lines)
-  filter!(x -> ~ismatch(r"^\s*//", x), lines)
+  filter!(x -> ~ismatch(r"^\s*\/\/", x), lines)
+  filter!(x -> ~ismatch(r"^\s*\/\*", x), lines)
 
-  # Remove lines with comments after text
+
+  # Remove end-of-line comments
   for ln = 1:length(lines)
-    lines[ln] = split(lines[ln], r"%.*$")[1]
+    lines[ln] = split(lines[ln], r"%.*$")[1]    # %  comments
+    lines[ln] = split(lines[ln], r"\/\/.*$")[1]  # \\ comments
+    lines[ln] = split(lines[ln], r"/\*.*$")[1]   # \* comments
   end
 
-  # Remove lines that are empty
-  filter!(x -> (x != "\n"), lines)
-
-  # Remove new lines
-  for ln = 1:length(lines)
-    lines[ln] = strip(lines[ln])
-  end
 
   # Smoosh text back together
   text = join(lines)
 
+  # Remove new lines, spaces, carriage returns,
+  text = replace(text, "\t", " ")
+  text = replace(text, "\r", " ")
+  text = replace(text, "\n", " ")
 
   # Get variable names
   tmp = match(r"var (.*?);", text)
   variables = split(tmp[1], " ")
+  filter!(x -> ismatch(r"\S", x), variables) # Remove lines if no chars present
 
   # Get exogenous variable names
   tmp = match(r"varexo (.*?);", text)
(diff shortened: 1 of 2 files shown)


test/modfile_parser.jl
```diff
@@ -0,0 +1,125 @@
...
+    lines[ln] = strip(lines[ln])
+  end
+
+  # Smoosh text back together
+  text = join(lines)
+
+
+  # Get variable names
+  tmp = match(r"var (.*?);", text)
+  variables = split(tmp[1], " ")
```
spencerlyon2: I think things would be better if you used the regex `,?\s*` to split variables (exon and parameters too). This will split on a optional comma followed by 0 or more spaces. 

I think this will solve your problem with `\t` that you've seen in the quest model

test/modfile_parser.jl
```diff
@@ -0,0 +1,125 @@
...
+      tmp = filter!(x-> (x != ""), split(tmp[2], ";"))
+      for ln = 1:length(tmp)
+        shock_val = match(r"=(.*)", tmp[ln])[1]
+        push!(shocks, shock_val)
+      end
+    end
+  end
+
+
+  close(f)
```
spencerlyon2: Move the closing to be immediately after you are done using the file. Something like this:

```julia
f = open(__file_name__)
lines = readlines(f)
close(f)

# carry on
```

spencerlyon2: Better yet, use the method for open that accepts a function as the first argument:

```julia
lines = open(readlines, mod_file_name)
```

test/modfile_parser.jl
jagman88: I tried this, but using `split(tmp[1], r",?\s*")`  splits all characters (i.e. we get a vector of individual letters..), and `match(r",?\s*", tmp[1])` returns a single string with a space (i.e. `" "`).  I've left it as above, but if you can figure out what might work instead let me know.

Oh and the `\t` issue wasn't showing up at this stage. The `\t` seem to show up when the lines are joined together.

test/modfile_parser.jl
jagman88: Went with the first of these (just preference: seems a little more transparent to me).

jagman88:
Made some fixes. This can now *almost* entirely parse the `EA_QUEST3.mod` file. There are still several (small) issues that I need some help with:

* On lines 205-213 of the mod file, there is a section called "Specification of Modelbase Parameters". This is in the middle of what would otherwise be the calibration section. At the moment, the calibration section can't handle this exception. Can either of you think of something that will work? Right now, I've just commented this section out in the .mod file. 
* Line 175 of the mod file looks like this: `E_EX_R     =   1/BETAE-1;      -log(BETAE);`. This seems to be a mistake, as the `-log(BETAE);` isn't assigned to anything. Regardless, it poses a problem for parsing because the lack of assigning parameter throws an error. Right now I've commented out after the first semi-colon. Let me know if this is just a mistake in the mod file, or something else that we need to have an exception for.
* In the mod file, the `shocks` section does not assign values to the shocks in the same order as the shocks are declared in the `varexo` section earlier on in the file. This suggests to me that we should read the variance terms into a dictionary, and then pull them into a covariance matrix afterwards. This way, it is the dictionary that defines the order of the shocks, not the initial `exogenous` section. Any other suggestions to deal with this?
* There is an equation called `interest`, defined in the mod file from line 237, that is read in with a lot of extra white spaces when it is saved to the list. Can one of you suggest an easy way to eliminate this white space? (Sorry, my brain is super tired right now...). 

So with the two commented fixes I've applied (bullets one and two, above), `EA_QUEST3.mod` can be parsed on my end. Let me know once you guys test it.

spencerlyon2:
@jagman88 I'm turning to this now. Are the above issues still a problem or have any of them been resolved?

jagman88:
Bullets 1 and 2 are resolved as long as we are happy to keep the comments in place in the mod file. 
I've just addressed bullet 3 with the latest commit. 
I haven't touched bullet 4, since we don't need to eliminate spaces when using the equations. 

I've also added the data output dictionary, so it should come out in the right form now. Pending testing, it's all good to go!

albop:
@jagman88  convinced me the modfile, didn't comply with our specific syntax. So, it's ok to comment the faulty lines. Merging now.

Issue #22, opened 2016-07-18
dynare_import: comma separated variables are not recognized


albop:
In modfiles, variables can be declared in two ways:
- comma-separated as: https://github.com/EconForge/dolo/tree/master/examples/dynare_modfiles/example1.mod
- space-separated as: https://github.com/EconForge/dolo/tree/master/examples/dynare_modfiles/example2.mod

Currently only space-separation works. @jagman88 : do you want to extend the parser you wrote ?

spencerlyon2:
I believe this should be a fairly minor change to the regex used to split variable names.

albop:
We don't aim to import dynare files anymore.

PR #23, opened 2016-07-19
FIX: allow mod file parsing to handle comma separated variables

jagman88:
Quick fix for issue [#22](https://github.com/EconForge/Dolo.jl/issues/22). Let me know if you want/need something fancier than this.


spencerlyon2:
Works for now, thanks!

PR #24, opened 2016-09-16
Sl/unified model

albop:
I'd like to merge that one, given that this branch fixes some of the issues in master. Any objection ?


spencerlyon2:
Nope, let's merge. Even though it isn't totally polished, it will be the way forward so we might as well work from it

PR #26, opened 2016-09-17
Albop/exo fixes

albop:
@spencerlyon2 introduced a new way to define `exogenous:` processes in the branch dtcc, but it has a few problems:
   - it doesn't encompass multivariate processes well, for instance when they are correlated
   - we might want to define things differently, for instance with an exogenous transition equation
   - it introduces another semantic, where symbol names are treated differently when evaluating yaml structures. this could be done separately (see below).

This p.r. reverts to the old syntax for `distribution` and `discrete_transition`, while keeping the unifying changes. 

For the sake of clarity, exogenous is also moved back to `options:` in the yaml examples (the way data is interpreted in `options` is clearly defined). In the imported object, exogenous is still a primary field. I propose to keep it where it is in the yaml file, until a new satisfying syntax is found.

The objects defined in the yaml file, `MarkovChain`, `Normal` and `AR1`, now return multivariate processes, whether or not scalar arguments were supplied. This is consistent with the conventions so far. For `AR1`, an alias `VAR1` is introduced. Later on, we might want to restrict `AR1` to unidimensional processes, but so far, I don't see any cost in treating it as a multivariate process. Same for `Normal`.

I also implemented a very simple compatibility fix in `symbolic.jl`, so that DTMSCC models can be imported (and converted to DTCC models) without any change.

Regarding the changes to object syntax and language semantics, we should discuss and implement in dolang.jl first. If the function `eval_with` knows the symbols of the model, the knowledge can be used to create objects which use variables names as input. For instance, it would be possible to define an object `!MultivariateProcess` such that

```
exogenous: !MultivariateProcess
   z: !AR1(sigma=0.001)
   w: !AR1(sigma=0.001)
```
would be equivalent to:
```
exogenous: !VAR1
   sigma: [[0.001,0],[0,0.001]])
```
In this case (and also in the `!Domain` example we discussed a while ago, it seems sufficient to use this info for the keys, but I am not 100% sure it is a general feature. Let's discuss that separately.


spencerlyon2:
This is great, thanks. I'll take a closer look soon, but I did notice that the rbc_dtcc_mc.yaml file doesn't actually have a `!MarkovChain` tag (it appears to be commented out).

src/symbolic.jl
```diff
@@ -84,22 +84,29 @@ end
...
             error("Yaml file must define section $k for dtcc model")
         end
     end
 
     d = _symbol_dict(deepcopy(data))
     mt = pop!(d, :model_type, :dtcc)
-    Symbol(mt) == :dtcc || error("Only support dtcc models now")
-
+    Symbol(mt) == :dtcc || Symbol(mt) == :dtmscc || error("Only support dtcc and dtmscc models now")
+    if Symbol(mt) == :dtmscc
```
spencerlyon2: Should we check to make sure that the dtcc model has `exogenous` symbols?

albop: I think it is done below, on line 107.

albop:
Small incremental update: I reorganized the yaml files a little bit, and made the same quickfix to import dtcscc files.
Basically I think it can be used without any special recompilation, just by passing a dummy argument.

If we know (by checking incidence matrices, for instance) that shocks don't appear in t-1, we would just write:
```
m = yaml_import("examples/models/rbc_dtcc_iid.yaml")
s,e,x,p = m.calibration[:states,:exogenous,:controls,:parameters]
transition(m,nothing,s,x,e,p)
```

I tried to add `@inline _unpack_var(x::Void, i::Integer) =  nothing` in dolang, but it didn't do the trick. This can be fixed independently from this PR.

src/symbolic.jl
```diff
@@ -6,7 +6,7 @@ immutable SymbolicModel{ID} <: ASM{ID}
     symbols::OrderedDict{Symbol,Vector{Symbol}}
     equations::OrderedDict{Symbol,Vector{Expr}}
     calibration::OrderedDict{Symbol,Union{Expr,Symbol,Number}}
-    exogenous::OrderedDict{Symbol,Associative}
+    exogenous::Dict{Symbol,Any}
```
spencerlyon2: Do we want to keep this as OrderedDict, or is Dict ok?

src/symbolic.jl
```diff
@@ -84,22 +84,32 @@ end
...
+      d[:options][:exogenous] = pop!(d[:options],:distribution,nothing)
+    end
     recipe = RECIPES[:dtcc]
     nm = pop!(d, :name, "modeldoesnotwork")
     id = gensym(nm)
     options = pop!(d, :options, Dict{Symbol,Any}())
     defs = pop!(d, :definitions, Dict{Symbol,Any}())
-    exog = pop!(d, :exogenous, Dict{Symbol,Any}())
+    # exog = pop!(d, :exogenous, Dict{Symbol,Any}())
+    exog = get(options, :exogenous, Dict{Symbol,Any}())
```
spencerlyon2: I still don't think we are ensuring that when the model is not dtmscc or dtcscc (i.e. we have a dtcc) that we have exogenous symbols and options.

spencerlyon2: Sorry I see it now... it is in the error.

spencerlyon2:
Looks really good.

I'd ask that we keep consistent indentation (4 spaces) throughout the library. I think a few `if`/`else` blocks had what looked like 2 spaces.

PR #29, opened 2016-12-20
ADD: documentation to algorithms

jagman88:
Added documentation for:

- steady_state.jl
- time_iteration_direct.jl
- time_iteration.jl
- value_iteration.jl

One quick question I have about function names: inside value_iteration.jl, there is a function `evaluate_policy`, which actually seems to be evaluating and returning the value function for a given decision rule. Could we possibly have a better name that `evaluate_policy`, since that's not actually what it's doing? No worries if not, just need to make the documentation explicit. 

Also, @albop in the [roadmap](https://github.com/EconForge/Dolo.jl/blob/master/ROADMAP.md), you mention that you want language independent documentation. At the moment, I have language dependent arguments, returns, and so on. Could you let me know what style you might want these in? E.g. what would the following become?:  `x::Vector{Float64}`: Current control variables

### DIFF
--- a/src/algos/value_iteration.jl
+++ b/src/algos/value_iteration.jl
@@ -1,5 +1,17 @@
 using Optim
 
+
+"""
+#TODO: Is evaluate_policy the best name for this? Seems that we are evaluating the value function, or evaluating a current decision rule.
+
+Evaluate the value function under the given decision rule, `dr` (i.e. evaluating under the current guess for the policy function). Then, using the evaluated value function, construct a new interpolation object for the value function.
+
+# Arguments
+* `model::NumericModel`: Model object that describes the current model environment.
+* `dr::`: Current guess for the decision rule.
+# Returns
+* `drv::`: Value function.
+"""
 function evaluate_policy(model, dr; verbose=true, maxit=100, )
 
     β = model.calibration.flat[:beta]
@@ -91,7 +103,23 @@ function evaluate_policy(model, dr; verbose=true, maxit=100, )
 end
 
 
+"""
+Evaluate a value function object `drv` at a particular combination of values for states, controls, and exogenous variable(s).
+
+# Arguments
+* `model::NumericModel`: Model object that describes the current model environment.
+* `β::Float64`: Value of the discount factor.
+* `dprocess::`: Discretized exogenous process.
+* `drv::`: Current guess for the value function object.
+* `i::Int64`: Index of node for exogeous variable(s).
+* `s::Vector{Float64}`: Current state variables.
+* `x::Vector{Float64}`: Current control variables.
+* `p::Vector{Float64}`: Model parameters.
+# Returns
+* `E_V::`: Evaluated value function.
+"""
 function update_value(model, β::Float64, dprocess, drv, i, s::Vector{Float64}, x0::Vector{Float64}, p::Vector{Float64})
+
     m = node(dprocess,i)  ::Vector{Float64}
     E_V = 0.0
     for j=1:n_inodes(dprocess,i)
@@ -105,6 +133,17 @@ function update_value(model, β::Float64, dprocess, drv, i, s::Vector{Float64}, 
     return E_V
 end
 
+
+"""
+Solve for the value function and associated decision rule using value function iteration.
+
+# Arguments
+* `model::NumericModel`: Model object that describes the current model environment.
+* `dr::`: Initial guess for the decision rule.
+# Returns
+* `dr::`: Solved decision rule object.
+* `drv::`: Solved value function object.
+"""
 function solve_policy(model, dr; verbose=true, maxit=5000, )
 
     β = model.calibration.flat[:beta]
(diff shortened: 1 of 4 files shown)


src/algos/value_iteration.jl
```diff
@@ -1,5 +1,17 @@
 using Optim
 
+
+"""
+#TODO: Is evaluate_policy the best name for this? Seems that we are evaluating the value function, or evaluating a current decision rule.
```
spencerlyon2: The function name makes sense to me.

We are finding the value associated with a particular policy, or in other words, evaluating the policy.

jagman88: But it's quite literally an evaluation of the value function, not an evaluation of the policy (which would be something like `dr(states, shocks, ...)`). 

I'm also confused by the fact that there another function in the module called `update_value`. Since both functions are referring to the value function, and both are evaluating that value function in some way, why would one refer to a "policy" and the other refer to a "value"?

albop:
Thank you @jagman88 for this work ! I have done a quick review.
General questions first:
- for alternative implementations with default arguments there are strings like 
```
"""
Computes a global solution for a model via backward time iteration.
The time iteration is applied to the residuals of the arbitrage equations.

If the stochastic process for the model is not explicitly provided, the process is taken from the default provided by the model object, `model.exogenous`
"""
```
whose first part seems a bit redundant to me (but I like to fact that we say exactly where the default comes from). Is it standard Julia practice ?

- why `dr::` instead of `dr` when we don't specify the types ?
- I wonder if we should had some special types like `ListOfPoints` or `ListOfListOfPoints` instead of `Array{Array{Float64,2},1}` or alike. The reason would be that we could use the exact same type names in python, and also, that we would not guarantee a given data layout (the one we have now is probably suboptimal).

As for small details:
- in residual: `s` is a list of points, i.e. a matrix, `res` is the same as `x` (i.e. a list of list of controls on the grid, one for each value of the exogenous shock).
- in time_iteration_direct: you didn't comment `maxit`, `verbose`, and `tol`... which is good since we don't know the right options for now.
- `steady_state`: you don't give the type of the input/output dictionaries. More importantly the function names are really misleading: we are not computing steady-states, just using the calibrated values. Maybe it would be clearer to should be call them residuals`.
- `evaluate_policy`: I would keep just "Evaluate the value function under the given decision rule". What is returned should be made clear by the return type.
- `update_value` not sure it exactly matches the main docstring. It is basically a one step policy evaluation, or the evaluation of the right hand side of a bellman for a given choice of control. This is basically used to construct the function with computes the maximum. It is not a user facing function, but it might help to rename it bellman somehting (bellman_rhs ?)
- `solve_policy`: docstring is fine. Function name doesn't make sense except by comparison with `evaluate_policy`. Should be `value_function_iteration` or something more explicit.
```

jagman88:
Great, thanks for this. Responses in order:

1) I don't think this is standard, no. I just thought I'd copy across the basic information about the function in case someone got to the documentation from the alternative implementation, rather than the original function. We could delete these strings, but keep the default options with the notes to the original function? 

2) Oh, sorry, I left these open because I hadn't figured out exactly what types these were. They're probably not needed for decision rules and models, so let me delete those. 

3) Agreed. I'll put something like `ListOfLists` in for now as a place-holder. 

4) Ok, will fix that. 

5) Ok.

6) Let me look at this one. In general, though, if function names are misleading, I really think they need to be changed since many of these are user-facing functions. Maybe something like `steady_state_residuals`? 

7) Ok.

8) Let me review this one.

albop:
1. yes, I would tend to keep only what is specific to a given signature. Think of what is printed when you do `?function` . You get all methods.
3. I'm glad you agree, but me, I'm still wondering ;-) The drawback of adding such type aliases is they need to be documented somewehere else which adds an additional layer.
6. `steady_state_residuals`  is still misleading (and shouldn't it be 0? ) . The values used are the calibrated values, given in the calibration section. They don't necessarly represent the steady-state. I think `residuals(model, calibration)` is my preferred option.

jagman88:
Ok, I've done most of these, but a couple of further things:

1. You said: "steady_state: you don't give the type of the input/output dictionaries." Do you mean that I don't need to give the types? Currently they are listed as `OrderedDict` and `Dict`. Should I remove those? 

2. I've left `ListOfLists` as a place-holder for now, while you decide what you like :-)

albop:
@jagman88 : I'm (finally) merging your docstrings. Thanks.
They will need to be updated at some point to reflect the optional arguments once, we have made all super consistent. For instance, all solution method have an `infos::Bool` option. If true, a result structure is returned with some information about the solution.
Also, related to what you did this summer, the `solve_policy` function has a `maxit_howard` argument (for now), which controls how many recursive evaluations are made in each howard improvement step.

sglyon:
Did we mean to close or merge this one?

albop:
Merge. It's harmless. What did I press ?

sglyon:
Harmless, but needed! Thanks @jagman88

PR #30, opened 2017-01-12
ENH: replace evaluate function with methods for call

sglyon:
This Pr replaces the `evaluate` function with new methods for `Base.call`. So now instead of `evaluate(dr, args...)` we do `dr(args...)`


albop:
@sglyon : Any reason to not do it ?

sglyon:
nope.

Performance is the same, syntax is cleaner. No hesitation from me

albop:
OK then. Merging now.

sglyon:
Sounds good to me, did you mean to merge or close?

albop:
I clicked on both... Let me try again

PR #31, opened 2017-01-12
Sl/cleanup

sglyon:
This is a noisy PR, but it actually isn't that big.

The only substantial changes I made are:

- rename `set_values` to `set_values!`
- in `evaluate_policy` allow the user to pass the discount factor as an argument incase it isn't named beta
- in `evaluate_policy`  evaluate all the `felicity` functions instead of just the first one

The rest is mostly aesthetic (stuff like spaces after commas), some performance tweaks (don't allocate so much), and making more functions type stable.

### DIFF
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -7,6 +7,7 @@ using DataStructures: OrderedDict
 using YAML: load_file, load
 using Requests: get
 using NLsolve
+using Optim
 using QuantEcon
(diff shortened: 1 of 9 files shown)


albop:
Thank you very much. `set_values!` makes perfect sense.

The thing with felicity is we don't know what to do if it is vector valued.
I'll still merge it as is.

Issue #28, opened 2016-11-22
Roadmap for Julia push


albop:
@anasza, @jagman88, @spencerlyon2 

I have just merged into master, our work with @anasza. There is *one* implementation of `time_iteration_direct`, `time_iteration`, `value_iteration` which all work for any type of shock, thanks to new decision rules objects (look in `test/test_algos.jl` ) . Modulo a few possible improvements, it feels just like the right approach. I have listed below the pieces we still need to implement to get a fully working version: (more details in https://github.com/EconForge/Dolo.jl/wiki/ROADMAP)

- [ ] Simulate methods for all processes
- [ ] Simulate a model with a solution
- [ ] AR1 decision rule
- Solution algorithms
    - [ ] VFI (cleanup)
    - [ ] PEA
    - [ ] perturbation (requires fix to AR1)
- [ ] Distribution object
- Ergodic distribution
    - [ ] Simulation based
    - [ ] By finding the fixed point of a discrete Markov Chain.
- [ ] Accuracy measures

spencerlyon2:
This is really cool! I don't think the merged master branch has been pushed to github yet, but I checked it out on the other branch.

I don't have the time to go through it carefully right now, but I will be doing that at some point.

albop:
Indeed, I didn't push the merged version. Done.

PR #32, opened 2017-03-02
Rewrite of decision rules + AR1 models.

albop:
Here is a PR for discussion (it will be merged automatically after a few days).

Off the top of my head, here is a list of changes (1 and 2 are where it is important to get design right)
- process.jl : a discretized object is now an object with a grid of states today (in grid.jl either empty, one point, cartesian, smolyak, unstructured) and a set of integration points/weights starting from each state
- a decision rule object is essentially a function `dr(x,y)` which holds the values defined on two different grids for x (exogenous) and y (endogenous), and knows how to interpolate them. To unify the solution methods, and potentially leave room to some optimizations, there is a CachedDecisionRule object which holds a decision rule, a discretized process and knows how to compute `dr(i,j,x)`
- models with an exogenous AR1 process can now be solved
- perturbation algorithm is extended to deal with non IID exogenous processes. It still uses ForwardDiff.jl 

Examples in test_algo.jl should run without problem


albop:
I updated roadmap https://github.com/EconForge/Dolo.jl/issues/28 accordingly.

albop:
I shall also say I am not 100% happy with the decision rules as they are now. In particular, it might make sense to have objects `DecisionRule{CartesianGrid, CartesianGrid, Linear}` where one specifies the kind of interpolation. Not sure whether this is better or worse than `LinearDecisionRule{CartesianGrid, CartesianGrid}`

src/numeric/processes.jl
```diff
@@ -8,36 +10,62 @@ abstract DiscreteProcess <: AbstractProcess
...
+iweight(dp::DiscretizedProcess, i::Int, j::Int) = dp.integration_weights[i][j]
+
+# date-t grid is unstructured
+type DiscreteMarkovProcess <: AbstractDiscretizedProcess
+    grid::UnstructuredGrid
+    transitions::Matrix{Float64}
+    values::Matrix{Float64}
+end
+
+DiscreteMarkovProcess(transitions::Matrix{Float64}, values::Matrix{Float64}) =
```
sglyon: It is very likely that I just don't understand what the fields of this type do, but don't you want `grid` to be the same as `values` instead of the same as `transitions`?

If I were to guess, I'd say that `grid` is supposed to be time t values, `values` is supposed to be time t+1 values, and `transitions` is the probability matrix the links the two.

src/numeric/decision_rules.jl
```diff
@@ -1,9 +1,43 @@
...
-abstract AbstractDecisionRule
-abstract AbtsractCubicSplineDR <: AbstractDecisionRule
+# abstract AbstractCachedDecisionRule{S,T} <: AbstractDecisionRule{S,T}
+# we don't implement that using subtypes anymore
+
+type DecisionRule{S,T} <: AbstractDecisionRule{S,T}
+    grid_exo::S
+    grid_endo::T
+    n_x::Int # number of values (could be a method if there was a nice method name)
+    coefficients::Array{Array{Float64}}
```
sglyon: Do we know the dimensionality of the inner Arrays here? If so we should add that info so the compiler knows what's going on

src/numeric/decision_rules.jl
```diff
@@ -1,9 +1,43 @@
...
+end
+
+
+type ConstantDecisionRule
+    constants::Vector{Float64}
+end
+
+(dr::ConstantDecisionRule)(x::Vector{Float64}) = dr.constants
+(dr::ConstantDecisionRule)(x::Matrix{Float64}) = repmat(dr.constants',size(x,1),1)
```
sglyon: What is the interpretation of the methods `dr(x)`? Is this evaluating the decision rule at the passed `x` points on the exogenous grid and all the `y` points on the endogenous grid?

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
+#####
 
-function set_values!(dr::AbtsractCubicSplineDR, values::Array{Array{Float64,2},1})
-    grid = dr.grid
-    a = grid.min
-    b = grid.max
-    orders = grid.n
-    dr.coefficients = [filter_mcoeffs(a, b, orders, v) for v in values]
-end
+using splines
```
sglyon: Can we do this once in the file `Dolo.jl`?

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
 
-function set_values!(dr::AbtsractCubicSplineDR, values::Array{Float64,2})
-    set_values!(dr, [values])
+function DecisionRule(grid_exo::EmptyGrid, grid_endo::CartesianGrid, n_x::Int)
+    orders = grid_endo.n
+    coeffs = [zeros(n_x, (orders+2)...)]
+    return DecisionRule{EmptyGrid, CartesianGrid}(grid_exo, grid_endo, n_x, coeffs)
```
sglyon: Specifying the type parameters here should be unnecessary. They will be inferred by the (implicitly defined) outer constructor. 

See how it worked a few lines below

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
-function DecisionRule(proc::MvNormal, grid::CartesianGrid, n_x::Int)
-    a = grid.min
-    b = grid.max
-    orders = grid.n
+function DecisionRule(grid_exo::CartesianGrid, grid_endo::CartesianGrid, n_x::Int)
+    # hmm kind of silently assuming we have cartesian grid
+    orders = grid_exo.n + grid_endo.n
     coeffs = [zeros(n_x, (orders+2)...)]
-    return IDecisionRule(proc, grid, n_x,  coeffs)
+    return (DecisionRule{CartesianGrid, CartesianGrid})(grid_exo, grid_endo, n_x, coeffs)
```
sglyon: Specifying the type parameters here is also not needed.

This should be written

```julia
DecisionRule(grid_exo, grid_endo, n_x, coefs)
```

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
-    b = grid.max
-    orders = grid.n
-    n_ms = size(mc.values, 1)
-    coeffs = [zeros(n_x, (orders+2)...) for v in 1:n_ms]
-    return MCDecisionRule(mc, grid, n_x,  coeffs)
+function set_values!(dr::AbstractDecisionRule{EmptyGrid, CartesianGrid}, values::Array{Array{Float64,2},1})
+    a = dr.grid_endo.min
+    b = dr.grid_endo.max
+    orders = dr.grid_endo.n
+    dr.coefficients = [filter_mcoeffs(a, b, orders, v) for v in values]
```
sglyon: Does splines.jl have filtering methods that don't allocate the coefficient vector? If so, given that we allocate memory for the coefficients in the constructor (when we call `zeros`), it would be great to use the memory here.

The same comments holds to other `set_values!` methods

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
+    set_values!(dr, [values])
 end
 
-function (dr::MCDecisionRule)(i::Int, x::Array{Float64,2})
-    a = dr.grid.min
-    b = dr.grid.max
-    n = dr.grid.n
-    cc = dr.coefficients[i]
-    res = splines.eval_UC_multi_spline(a, b, n, cc, x)'
+function evaluate(dr::AbstractDecisionRule{EmptyGrid, CartesianGrid},z::Matrix{Float64})
```
sglyon: Is calling this `evaluate` so we can dispatch on the abstract type?

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
-function DecisionRule(proc::MvNormal, grid::CartesianGrid, values::Array{Float64,2})
-    return DecisionRule(proc, grid, [values])
+function set_values!(dr::Dolo.AbstractDecisionRule{Dolo.CartesianGrid, Dolo.CartesianGrid}, values::Array{Array{Float64,2},1})
+    a = cat(1,dr.grid_exo.min,dr.grid_endo.min)
+    b = cat(1,dr.grid_exo.max,dr.grid_endo.max)
+    orders = cat(1,dr.grid_exo.n,dr.grid_endo.n)
+    n_x = size(values[1],2)
+    vv = zeros(prod(dr.grid_exo.n), prod(dr.grid_endo.n), n_x)
+    for n=1:size(vv,1)
+        # ah, if only Julia had chosen C-order arrays !
```
sglyon: If only...

src/numeric/decision_rules.jl
```diff
@@ -17,190 +51,169 @@ function filter_mcoeffs(a::Array{Float64,1},b::Array{Float64,1},n::Array{Int64,1
...
-    cc = dr.coefficients[1]
-    N = size(x, 1)
-    e = dr.process.integration_nodes[i][j, :]
-    z = cat(2, repmat(e', N, 1), x)
-    res = splines.eval_UC_multi_spline(a, b, n, cc, z)'
-    return res
+
+
+#####
+##### Cached Decision Rules (do we really need them ?)
```
sglyon: Am I right that the `CachedDecisionRule` is just a placeholder for now?

If so, that is ok -- I just wanted to make sure I didn't miss something super clever/subtle

src/numeric/grids.jl
```diff
@@ -5,19 +5,48 @@ function mlinspace(min, max, n)
     return QE.gridmake(nodes...)
 end
 
-type CartesianGrid <: Grid
```
sglyon: This comment applies to the function `mlinspace` above (it wasn't changed in this PR, so I can't comment on it...)

You can define `nodes = map(linspace, min, max, n)`, which I prefer to the comprehension and splatting from a style perspective

src/numeric/grids.jl
```diff
@@ -5,19 +5,48 @@ function mlinspace(min, max, n)
...
-    min::Array{Float64,1}
-    max::Array{Float64,1}
-    n::Array{Int64,1}
+type EmptyGrid <: Grid
```
sglyon: Let's make this immutable

src/numeric/grids.jl
```diff
@@ -5,19 +5,48 @@ function mlinspace(min, max, n)
...
 
-type SmolyakGrid <: Grid
-    min::Array{Real,1}
-    max::Array{Real,1}
-    mu::Array{Int,1}
+
+type PointGrid <: Grid
+    point::Vector{Float64}
+end
+nodes(grid::PointGrid) = grid.point[:,:]
```
sglyon: From reading this code, it seems that the other types of grids have that each row is a node across all variables and each column is all nodes for a variable. 

If that is the case, this method is wrong. 

We should define `nodes(grid::PointGrid) = grid.point'` so the output is `(1, n)`

src/numeric/grids.jl
```diff
@@ -5,19 +5,48 @@ function mlinspace(min, max, n)
...
+    function CartesianGrid(min, max, n)
+        nodes = mlinspace(min, max, n)
+        return new(min,max,n,nodes)
+    end
+end
+nodes(grid::Grid) = grid.nodes
+n_nodes(grid::Grid) = size(grid.nodes,1)
+node(grid::Grid, i::Int64) = grid.nodes[i,:]
+
+type SmolyakGrid <: Grid
```
sglyon: Cleaning up the Julia Smolyak code and merging into BasisMatrices has been on my todo list for quite some time.

When it becomes blocking over here please let me know and that will provide extra incentive to clean it up.

src/numeric/processes.jl
```diff
@@ -1,5 +1,7 @@
 import QuantEcon
 const QE = QuantEcon
+import Distributions:rand
```
albop: Sure.

src/numeric/processes.jl
```diff
@@ -8,36 +10,62 @@ abstract DiscreteProcess <: AbstractProcess
 abstract ContinuousProcess <: AbstractProcess
 abstract IIDExogenous <: AbstractProcess
 
-abstract AbstractDiscretizedProcess <: DiscreteProcess
+abstract AbstractDiscretizedProcess # <: DiscreteProcess
```
albop: That's a no brainer. Let's say we are aiming for 1.0 compatibility ;-)

src/numeric/processes.jl
```diff
@@ -8,36 +10,62 @@ abstract DiscreteProcess <: AbstractProcess
...
+###
 
-# discretized process
-
+# date-t grid has a known structure
 type DiscretizedProcess <: AbstractDiscretizedProcess
-    nodes::Matrix{Float64}
-    integration_nodes::Array{Matrix{Float64}}
-    integration_weights::Array{Vector{Float64}}
+    grid::Grid
```
albop: OK

src/numeric/processes.jl
```diff
@@ -8,36 +10,62 @@ abstract DiscreteProcess <: AbstractProcess
...
 end
 
-n_nodes(dp::DiscretizedProcess) = size(dp.nodes, 1)
+n_nodes(dp::DiscretizedProcess) = n_nodes(dp.grid)
+node(dp::DiscretizedProcess, i) = node(dp.grid,i)
 n_inodes(dp::DiscretizedProcess, i::Int) = size(dp.integration_nodes[i], 1)
-inodes(dp::DiscretizedProcess, i::Int, j::Int) = dp.integration_nodes[i][j, :]
-iweights(dp::DiscretizedProcess, i::Int, j::Int) = dp.integration_weights[i][j]
-node(dp::DiscretizedProcess, i) = dp.nodes[i, :]
+inode(dp::DiscretizedProcess, i::Int, j::Int) = dp.integration_nodes[i][j, :]
```
albop: So this is the big elephant in the room.
In general terms, it is quite clear that the current choices of data layout are far from optimal. As you correctly note, the strategy is to isolate algorithmic code as much as possible from this decision, hence the functions `inode`, `iweight`, ... The most problematic point right now, is the storage of the computed values or the residuals. Currently it is `Array{Array{Float64,2},1}` which is a half-baked solution. There is also no abstraction on top of it to hide this choice.
As for the data layout of DiscretizedProcess I would advocate instead `dp.integration_nodes[i][j]` to get a vector. Indeed, we don't make anywhere the assumption that the number of integration nodes has to be the same for all states today, which allows for the design of fancy markov process.

src/numeric/processes.jl
albop: Good catch, that's a typo. In the case of Markov chains, the nodes attainable today and tomorrow are the same, and they are given by `values`

src/numeric/processes.jl
```diff
@@ -8,36 +10,62 @@ abstract DiscreteProcess <: AbstractProcess
...
+inode(dp::DiscretizedIIDProcess, i::Int, j::Int) = dp.integration_nodes[j, :]
+iweight(dp::DiscretizedIIDProcess, i::Int, j::Int) = dp.integration_weights[j]
 node(dip::DiscretizedIIDProcess, i) = zeros(n_inodes(dip, 1))
 
+
+
+
 # Normal law
 
 immutable MvNormal <: IIDExogenous
```
albop: What do you mean by wrap ?
Subtyping Distributions.MvNormal ? Or synchronize the fields of Dolo.MvNormal ? Some kind of multiple inheritance (which apparently won't make it in to 1.0) ?
In principle I'm not against using the Distributions object rather directly, rather the opposite, but I was attempting to proceed cautiously. There are small details I'm not completely sure about. Like do we officially require users to use unicode symbols in yaml file ? In some environments it is quite a pain...

src/numeric/processes.jl
```diff
@@ -55,22 +83,31 @@ function discretize(mvn::MvNormal)
     DiscretizedIIDProcess(x, w)
 end
 
-
-# discrete markov processes
-
-type DiscreteMarkovProcess <: DiscreteProcess
-    transitions::Matrix{Float64}
-    values::Matrix{Float64}
+function rand(mvn::MvNormal, args...)
```
albop: Funnily enough, cholesky is the reason why I didn't want to use the Distributions object directly. Reason for that is that if you do `Distributions.MvNormal(zeros(2), zeros(2,2))` (one diagonal zero is enough) you get an error. But we absolutely do want to be able to deal with covariance matrices containing zero eigenvalues, as it comes up quite often. Otherwise, you constantly find yourself defining things like `sigma = 0.000000001` to see what si the perfect foresight version of your model.
Now, if you tell me what kind of wrapper you have in mind, I might understand better the slackness we have to extend/modify the Distributions objects.

src/numeric/processes.jl
```diff
@@ -110,116 +129,59 @@ end
...
-        integration_nodes = collect(yⁱʲ[:, :, ii] for ii in 1:n_states[n])
-
-        # Get transition probabilities
-
-        integration_weights = w[n]
-        integration_weights =[integration_weights]
-      end
-#  return nodes, integration_nodes, integration_weights
-    return DiscretizedProcess(nodes, integration_nodes, integration_weights)
+function discretize(var::VAR1, n_states::Array{Int64,1}, n_integration::Array{Int64,1}; n_std::Int64=2,)
```
albop: Are you sure you don't want to become a Debian maintainer ? ;-)
Joke aside, my confusion came from the day I tried to replace all Float64/Int64 by AbstractFloat/Integer. It broke everything, and left me very frustrated so I came back to the most concrete type I know of.
But, the transition Int64->Int seems harmless to me.

src/numeric/processes.jl
```diff
@@ -110,116 +129,59 @@ end
...
+
+    R = var.R
+    M = var.M
+    Sigma = var.Sigma
+    S = QE.solve_discrete_lyapunov(R,Sigma)  # asymptotic variance
```
albop: yeah, that one also fails for degenerate Sigma, so I have also considered rewriting it. Maybe I'll just open an issue against QuantEcon.

src/numeric/processes.jl
```diff
@@ -110,116 +129,59 @@ end
...
+    #- E - simulated innovations, ordered as [N, T, n], where n is the number of variables;
   """
-  n = size(VAR_info.M, 1);
-  srand(123) # Setting the seed
-  d = MvNormal(VAR_info.Sigma);
+  n = size(var.M, 1);
+  # srand(123) # Setting the seed
+  d = MvNormal(var.Sigma);
   XN=zeros(N, T, n);
   E=zeros(N, T, n);
```
albop: I haven't touched much these simulation routines, so I'll defer to @AnasZa . I don't remember whether we needed the innovations E or not, and for what. I guess there are some cases where we do.

src/numeric/processes.jl
```diff
@@ -110,116 +129,59 @@ end
...
   #Inicial conditions
   XN[:, 1, :]=rand(N, n);
   for jj = 1:1:N
         E[jj, :, :] = rand(d, T)';
         for ii =1:1:T-1
-        XN[jj, ii+1, :]=VAR_info.M+VAR_info.R*(XN[jj, ii, :]-VAR_info.M)+E[jj, ii, :];
+            XN[jj, ii+1, :]=var.M+var.R*(XN[jj, ii, :]-var.M)+E[jj, ii, :];
```
albop: Although you are correct about efficiency, let's not rush to change that one:
1/ it's currently not a bottleneck for anything
2/ we still don't know which is the actually the desired ordering.

src/numeric/decision_rules.jl
albop: We don't know generically but we know for a given model. The element `coefficients[1]` has dimension `n_x \time n_1 ... \times n_d`. This dimension is used to dispatch to the right filtering/interpolating functions. I guess you could parameterize everything by the number of dimensions (maybe parameterize Grid), but only if it makes a sizable performance gain.

src/numeric/decision_rules.jl
albop: That's an easy one. To paraphrase Gertrude Stein, "a constant is a constant is a constant" . So no matter where and how you evaluate it, it remains the same.
In general `dr(x)` makes sense for an iid driving process as there is no exogenous process in this case. Or when there is a continuous exognous process you can define without ambiguity `dr([x1;x2])` as `dr(x1,x2)`.

src/numeric/decision_rules.jl
albop: sure that will come. But I'm not a big fan of doing it early as it makes it harder to find the dependence of a particular piece of code, as opposed to the library as a whole.

src/numeric/decision_rules.jl
albop: All right

src/numeric/decision_rules.jl
albop: Yes it has, and yes, it *should* be used here, but so far no attention was devoted to perform inplace operations. In fact:
- I don't necessarly intend to keep splines.jl forever if we can do as good with BasisMatrices
- there is fundamental ordering problem (which shows in the `{Cartesian,Cartesian}` flavour which is that the ordering of the values is really ill-suited to be filtered by the product of the splines basis matrices. So we can expect further changes/discussions on how to store coefficients.

src/numeric/decision_rules.jl
albop: Yes. I introduced it when I was trying to subtype the decision rules, for the Cached version. You cant inherit the call method, so I inherited the evaluate method instead. But I couldn't find a sastisfying object design and reverted to the simple wrapper we have now. I don't know whether we will need inheritance again so I have kept `evaluate` just in case.

src/numeric/decision_rules.jl
albop: yeah, I'm still 20% tempted to use our own C-order arrays in some definite parts of the libraries...

src/numeric/decision_rules.jl
albop: So, despite what I initially thought, CachedDecisionRule won't be very useful, unless you have a large number of points for the discretized processes (I think you had such a use case at some point). But it is not what is required to make fast VFI when transitions are non-stochastic. (then you need to cache the states you transition too, and another object).
However, it is useful right now to give a precise meaning to `dr(i,x)` and `dr(i,j,x)` even for decision rules which are continuous for both arguments. So this allows to unify the solution code.
If you have a better name than CachedDecisionRule, I'll take it.

src/numeric/grids.jl
albop: OK, I'll update `mlinspace`.

src/numeric/grids.jl
albop: Sure.
But... here we are back to the question of what is the good way to add cached data to an object. In this case, it might be that constructing the grid is costly and you don't want to do it by default. But when you construct it you want to keep it. Then, hypothetically, what is the right way to proceed ? Subtype ? Create another nutable type with a grid field, a fillable nodes slot ?

src/numeric/grids.jl
albop: good catch

src/numeric/grids.jl
albop: Let's consider it as non blocking until we are happy with the overall design. As an interim measure, I guess we can still rely on some methods from Smolyak.jl or include them temporarly.
I would say it is still medium/medium-high priority.

albop:
@sglyon : thank you very much for the detailed review ! That was massively helpful.

src/numeric/decision_rules.jl
albop: So removing `{EmptyGrid, CartesianGrid}` doesn't worrk so I'm going to keep it until I understand what's going on...

src/numeric/processes.jl
sglyon: That would be cool to have `dp.integration_nodes[i][j]` return a vector. To get that to happen maybe we need an `IntegrationNodes` or `IntegrationWeightsNodes` type that we can define a custom `getindex` method for? I'm not sure right now...

Can you show me an example of that Vector of Matrices you talked about?

What do you suggest we do in this PR regarding  the non-optimal data layout in these routines? Should we merge as-is, write some tests, then revisit the storage order in a separate issue?

src/numeric/processes.jl
sglyon: So what I had in mind was something like this:

```julia
type MvNormal <: IIDExogenous
    dist::Distributions.MvNormal
end

MvNormal(Sigma::Matrix{Float64}) = Distributions.MvNormal(zeros(size(Sigma, 1)), Sigma)
MvNormal(sigma::Float64) = MvNormal(reshape([sigma], 1, 1))
get_mu(mvn::MvNormal) = mean(mvn.dist)
get_Sigma(mvn::MvNormal) = cov(mvn.dist)
Base.length(mvn::MvNormal) = length(get_mu(mvn))

function discretize(mvn::MvNormal)
    n = fill(5, length(mvn))
    x, w = QE.qnwnorm(n, get_mu(mvn), get_Sigma(mvn))
    DiscretizedIIDProcess(x, w)
end

rand(mvn::MvNormal, args...) = rand(mvn.dist, args...)

function simulate(mvn::MvNormal, n_exp::Integer, horizon::Integer)
    d = length(mvn)
    out = zeros(d, n_exp, horizon)
    for t=1:horizon
        out[:,:,t] = rand(mvn, n_exp)
    end
    return out
end
```

src/numeric/processes.jl
sglyon: Ok, that makes sense. I actually ran into this problem in QuantEcon and used some silly hack to get around it (see [here](https://github.com/QuantEcon/QuantEcon.jl/blob/cde1f395243151ce54f3f1ba6b7d079b751878d0/src/lss.jl#L26-L38)).

Given the updated comment about wrapping the distributions object, what do you think is the best way forward?

As of right now this method won't work as it is because cholesky doesn't accept a matrix of zeros, so whatever route we choose we still have to be clever in getting around that issue.

src/numeric/processes.jl
sglyon: An issue in QE would be great

src/numeric/processes.jl
sglyon: I totally agree that we don't need to rush and change all the C-ordered algorithms in Dolo.jl, but I think here it is a no-brainer to change it to be `(n, T, N)` the whole time.

src/numeric/decision_rules.jl
sglyon: Let's hold off for now

src/numeric/decision_rules.jl
sglyon: Ok, so the `x` here is observation the endogenous state?

src/numeric/decision_rules.jl
sglyon: Strange, as it seems to work a couple lines below. 

Oh well, let's do what works!

src/numeric/decision_rules.jl
sglyon: If I find some time later this week (and after this is merged) maybe I'll try to add a Decision rule type based on BasisMatrices and we'll see what the performance difference

src/numeric/decision_rules.jl
sglyon: Is this just a matter of being used to thinking about the algorithms in C-order, or is there something fundamental that requires it?

src/numeric/decision_rules.jl
sglyon: I don't have a great sense of how they are used (yet), so I can't think of any better names

src/numeric/processes.jl
albop: I don't think we need IntegrationNodes objects. An integration node (a vector) is a feature of a discretized process, so we could directly do `dprocess[i,j]` or somthing alike. But this is syntactic sugar, so let's not do that.
The vector of matrices ? Look at the `x` argument of `residual` in https://github.com/EconForge/Dolo.jl/blob/master/src/algos/time_iteration.jl#L1 .
That one is of type `Array{Array{Float64,2},1}` : for each value of the exogenous shock, we have a matrix of values on the endogenous grid. Memory-wise this correspond to an Array{Float64,3} with dimensions `N_endo.n_x.N_exo` which is useful here because, we perform the extraction `x[i] ~ x[:,:,i]` many times, but inadequate for a human being: we want to think of a `N_exo.N_endo.n_x` array. 
So this is not perfect and not to be addressed in this PR. Essentially, I would like to stabilize user-facing functions first before revisiting all the memory issues. The solution might be a big dimension reversion, or some more levels of abstractions. But let's be super cautious here as iterating back and forth is not an option.

src/numeric/processes.jl
albop: I'm not sure about this wrapper. It looks a bit weird, since it doesn't have the basic attributes you would expect (sigma, mu) and still doesn't work for semi-definite matrices.
Let's open issues and revisit later.

src/numeric/decision_rules.jl
albop: If you are defining decision rules on two grids, you evaluate dr(x,y). If you only one non-trivial grid, you evaluate on dr(x). In any case, it doesn't matter.

albop:
I'm merging that one and will make new modifications in a new branch.

PR #34, opened 2017-03-30
Azh exp

AnasZa:


examples/models/rbc_dtcc_iid_2ar1.yaml
```diff
@@ -0,0 +1,92 @@
...
+    i: delta*k
+    c: y - i
+    V: log(c)/(1-beta)
+    u: c^(1-sigma)/(1-sigma) - chi*n^(1+eta)/(1+eta)
+    m: exp(d)*beta/c^sigma*(1-delta+rk)
+
+options:
+
+    exogenous: !Normal
+        sigma: [[sig_z**2, 0], [0, sig_d**2]]
```
sglyon: I can't remember if we handle the `**` in Julia or if we require `^`.

Does it work as is?

src/algos/simulation.jl
```diff
@@ -20,49 +22,102 @@ function simulate(model::AbstractNumericModel, dr::AbstractDecisionRule,
...
+    for i in 1:N
       s_simul[i, :, 1] = s0
       x_simul[i, :, 1] = x0
     end
-    # NOTE: this will be empty if ny is zero. That's ok. Our call to `cat`  #       below will work either way  y_simul = Array(Float64, n_exp, ny, horizon)
+    # NOTE: this will be empty if ny is zero. That's ok. Our call to `cat`  #       below will work either way  y_simul = Array(Float64, N, ny, horizon)
 
-    for t in 1:horizon
+    for t in 1:T
         s = copy(view(s_simul, :, :, t))
```
sglyon: Here I think we should remove the `copy` as it defeats the purpose of calling `view`

AnasZa: Yeah, but that's not working without `copy`.

examples/models/rbc_dtcc_iid_2ar1.yaml
AnasZa: I changed for `^`, but it works either way.

sglyon: Oh ok, if it works I'm happy either way. I just wanted to make using `**` wasn't doing something strange.

src/algos/simulation.jl
sglyon: I can't see why that would happen. I'll try to experiment with the code and take a closer look

albop:
So judging from the code, the `copy(view(` issue doesn't come from this PR, but rather from the DecisionRule object (some abstractarray missing I guess). I'm opening a separate issue to track that one.
As for me, I'm rather happy with this PR. @sglyon : if you want some time to look at it, I can wait, otherwise I'll merge it right away.

I actually like the new conventions: `simulate(model, dr, ...)` always return a 3d array while `response(model, dr, :e_z)` or `response(model, dr, e1::vector)` always returns a return a 2d array.

On a side note: simulating models with markov chains will need to be addressed separately.

sglyon:
Let's open issues for the `copy(view(` issue as well as the Markov chain simulations.

albop:
And, I forgot, thank you very much @AnasZa  from working on that !

PR #35, opened 2017-03-30
FIX: Array->AbstractArray in DecisionRules

albop:
This complements the discussion in #34 . 

Problem was the evaluation of d.r. rules was defined like `dr(x::Array{Float64})` which makes it incompatible with array views. I haven't changed the other defintions though (meaning that coefficients of a d.r. still must be contiguous arrays).

I did another small fix: renamed the constant field of VAR1 objects from  `M` to `mu` which makes it consistent with the notation in `MvNormal` (and fixes the `response` function).


sglyon:
Thanks!

PR #45, opened 2017-05-05
Compatibility with 0.6

albop:
This is not ready to merge.

For now, there seem to be an incompatibility in Dolang:

```
julia> Dolo.yaml_import(Pkg.dir("Dolo", "examples", "models", "rbc_dtcc_mc.yaml"))
ERROR: ArgumentError: method definition for expectation! at :0 has free type variables
Stacktrace:
 [1] #NumericModel#78(::Bool, ::Type{T} where T, ::Dolo.SymbolicModel{Symbol("##Real Business Cycle#721")}) at /home/pablo/.julia/v0.6/Dolo/src/numeric.jl:75
 [2] (::Core.#kw#Type)(::Array{Any,1}, ::Type{Dolo.NumericModel}, ::Dolo.SymbolicModel{Symbol("##Real Business Cycle#721")}) at ./<missing>:0
 [3] yaml_import(::String) at /home/pablo/.julia/v0.6/Dolo/src/model_import.jl:37
```


src/numeric/decision_rules.jl
```diff
@@ -29,7 +29,7 @@ end
 (dr::ConstantDecisionRule)(i::Int, j::Int, x::Union{AbstractVector{Float64},AbstractMatrix{Float64}}) = dr(x)
 
 
-type BiTaylorExpansion <: AbstractDecisionRule
+@compat type BiTaylorExpansion <: AbstractDecisionRule{EmptyGrid,EmptyGrid}
```
sglyon: Here let's use the new syntax, which means either using `@compat struct` if `BiTaylorExpansion` is meant to be immutable or `@compat mutable struct` if it needs to be mutable.

PR #49, opened 2017-05-21
Albop/dict model

albop:
This branch does many different things:
1/ introduce a new, simpler model object, which breaks the syntax in the yaml files. Contrary to its Python version, it doesn't intend to be backward compatible. 
2/ merge the simulation work from @AnasZa 
It touches many different bits and moves a few things around. I'd like to merge it as it is so that every one can start working on the new version. We can make refinement later.
From now on, I'll try to make more atomistic PR ;-)


new_symbolic.jl
```diff
@@ -0,0 +1,163 @@
...
+
+pairs = [("!Cartesian", :Cartesian),
+         ("!Smolyak", :Smolyak),
+         ("!Normal", :Normal),
+         ("!MarkovChain", :MarkovChain),
+         ("!AR1", :AR1)]
+yaml_types = Dict{AbstractString,Function}([(t, (c, n) -> construct_type_map(s, c, n))
+                       for (t, s) in pairs])
+
+abstract type ASModel end
```
sglyon: `ASModel` is defined as and abstract type here. It is also defined in the Dolo.jl file

src/Dolo.jl
```diff
@@ -35,17 +30,26 @@ export yaml_import, eval_with, evaluate, evaluate!, model_type, name, filename,
...
+export ModelCalibration, FlatCalibration, GroupedCalibration
 
-typealias ASM AbstractSymbolicModel
-typealias ANM AbstractNumericModel
+# set up core types
+abstract AbstractSymbolicModel{ID}
+# abstract AbstractSymbolicModel{ID} <: AbstractModel{ID}
+abstract AbstractModel{ID} <: AbstractSymbolicModel{ID}
 
+typealias ASModel AbstractSymbolicModel
```
sglyon: ...Other `ASModel` definition

src/Dolo.jl
```diff
@@ -35,17 +30,26 @@ export yaml_import, eval_with, evaluate, evaluate!, model_type, name, filename,
...
-abstract AbstractSymbolicModel{ID} <: AbstractModel{ID}
-abstract AbstractNumericModel{ID} <: AbstractModel{ID}
+export ModelCalibration, FlatCalibration, GroupedCalibration
 
-typealias ASM AbstractSymbolicModel
-typealias ANM AbstractNumericModel
+# set up core types
+abstract AbstractSymbolicModel{ID}
+# abstract AbstractSymbolicModel{ID} <: AbstractModel{ID}
+abstract AbstractModel{ID} <: AbstractSymbolicModel{ID}
```
sglyon: What is the difference between `AbstractModel` and `AbstractSymbolicModel` now? Is one supposed to be a numerical object? If so isn't the old set of types where we are clear about symbolic vs numeric easier to understand?

test/model_types.jl
```diff
@@ -259,96 +260,96 @@
...
-        @test maxabs(v0 - v1) < 1e-13
-
-        bounds = ones(x0)
-        controls_ub!(m, bounds, e_, s0, p)
-        @test bounds == [Inf, Inf]
-
-        controls_lb!(m, bounds, e_, s0, p)
-        @test bounds == [0.0, 0.0]
-    end
+    #
```
sglyon: I realize we are trying to get this done quickly, but commenting out tests instead of adjusting them just feels like a bad idea

src/numeric/processes.jl
```diff
@@ -116,6 +116,58 @@ function response(mvn::MvNormal, e1::AbstractVector; T::Integer=40)
...
+            running = false
+        else
+            i += 1
+        end
+    end
+    return i
+end
+
+# function simulate(nodes::Array{Float64,2}, transitions::Array{Float64,2},N::Int, T::Int; i0::Int=1)
+function simulate(process::Dolo.DiscreteMarkovProcess,N::Int, T::Int, i0::Int; return_indexes=true)
```
sglyon: Should we leverage the QuantEcon Markov chain simulation routines here?

new_symbolic.jl
```diff
@@ -0,0 +1,163 @@
+import YAML
```
sglyon: What is the difference between this file and the src/model.jl?

It seems like many things are repeated between the two files. and that this file is not included in the package

test/test_tabulate.jl
```diff
@@ -0,0 +1,38 @@
+path = Pkg.dir("Dolo")
+
+Pkg.build("QuantEcon")
```
sglyon: We shouldn't have to do this inside the tests

test/test_tabulate.jl
```diff
@@ -0,0 +1,38 @@
...
+import Dolo
+using DataArrays
```
sglyon: I don't think we directly use `DataArrays`, `DataFrames`, RDatasets`, or `DataStructures` in this file -- we shouldn't import them as it might cause issues if we decide to have CI run these tests.

new_symbolic.jl
albop: new_symbolic.jl should disappear.

src/Dolo.jl
albop: Now we don't have a NumericalModel type. Just Model. So AbstractNumericModel->AbstractModel.
SymbolicModel is just a thin layer to that data. We could call it smthg elsr

test/model_types.jl
albop: It looks like they are an easy task for one of the new joiners due to arrive this week and next week.

src/numeric/processes.jl
albop: Possibly but this is super simple so there isn't much to leverage.

test/test_tabulate.jl
albop: sure

test/test_tabulate.jl
albop: of course

src/Dolo.jl
sglyon: Ahh ok I see now, thanks. That makes more sense. 

So am I correct in saying we have 

- a new symbolic model type (`SModel`) that is just a wrapper around the Dict we get when we have YAML.jl parse the model file
- A `Model` type that includes the same underlying data (actually I had a question, is the `data` field on `Model` supposed to be a Dict or a `SModel`?) and numerical objects (some just plain Julia datatypes for now) for things like the grids, domain, etc.

src/util.jl
```diff
@@ -85,3 +85,27 @@ function solve_triangular_system(d::OrderedDict)
 
     OrderedDict{Symbol, Real}(zip(nms, data))
 end
+
+
+
+appenddim(A) = reshape(A, size(A)..., 1)
+
+# does kronecker product of matrices according to fortran order
+function fkron(A::AbstractMatrix{Float64}, B::AbstractMatrix{Float64})
```
sglyon: Isn't this just `kron(B, A)`?

```
julia> A = rand(2,2);

julia> B = rand(3,3);

julia> fkron(A,B) - kron(B,A)
6×6 Array{Float64,2}:
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0
```

src/algos/time_iteration.jl
```diff
@@ -103,11 +103,11 @@ If the stochastic process for the model is not explicitly provided, the process
 * `dr`: Solved decision rule.
 """
 function time_iteration(model, process, init_dr;
-      verbose::Bool=true, maxit::Int=100, tol::Float64=1e-8, infos::Bool=false)
+      verbose::Bool=true, maxit::Int=100, tol::Float64=1e-8, details::Bool=false)
 
     # get grid for endogenous
-    gg = model.options.grid
-    grid = CartesianGrid(gg.a, gg.b, gg.orders)  # temporary compatibility
+    grid = model.grid
```
sglyon: One idea here...

Should we have `get_grid(::Model)` like we have `get_grid(::SModel`), or do you prefer accessing the grid field?

The reason I ask is that it might make things easier to do "just in time" field updating (e.g. if the calibration has changed) if we go through a method instead of accessing the field itself.

albop:
- So the SModel is indeed a wrapper around the dict, whose goal is to read
and interpret the data through the `get_*` methods. They all return fully
functional objects so that calling it a `symbolic` model is a bit abusive.
- The user doesn't need to know about another model class `SModel`. But he
can inspect the data field which is made of simple objects.
- Basically the sole reason to exist for SModel, is to separate the coding
of the get_functions. Essentially it could be fused with the Model (which
derives from it anyway).

src/util.jl
albop: Yes indeed ;-)
(as long as operations are commutative...)

src/algos/time_iteration.jl
albop: Yes, that's what we do on the python side. There is a `model.get_grid(**additional_args_which_override_the_content_from_yaml)` .
I guess it would be enough to extend the current `get_*` method (doesn't need to be redef. for Model) but I just want to know how to do it systematically for nested structures (for instance, if you want to change the domain while getting the grid).

albop:
Ok. I merged that one. I'll open a separate issue to develop more systematic tests, especially for the solution routines.
Currently Travis fails, because solutions methods take too much time to return.

sglyon:
Practice that button pushing!

albop:
Come on, let’s get to it. I know that we can do it !

PR #50, opened 2017-05-23
– Added AxisArrays to

AnasZa:
- simulate(MC)
- simulate(mvn)  - this one enumerates exogenuous process  but doesn’t name them since an input is `exogenuous.process` and it doesn’t contain Symbols.
- simulate(VAR)
- tabulate
I checked everything works by running test_mc_ar1.jl

– I also corrected some typos in tabulate function. Now, everything works.


albop:
Thank you. Travis complains about an unrelated error. So I'll merge it now.

PR #52, opened 2017-05-28
WIP: options

albop:
I would like to have a consistent way to pass options to all functions in a way that is consistent with yaml file.
Let assume the yaml file ends up with:
```
domain:
   k: [0,1]
exogenous: !VAR1
   rho: sig_z
options:
   grid: !Cartesian
      n: [10,10]
   interpolation: cubic
   discretize: # doesn't exist
      method: rouwenhorst
      mu: 3
```
I suggest to define:
```
time_iteration(model; options=Dict())
```
`options` would expect dict-of-dict input and would be merged recursively with the data from the yaml file. For instance, one could pass `options=Dict(:grid=>Dict(:type=>:Smolyak))` and that would be equivalent to:
```
options:
   grid: !Smolak
      n: [10,10]
   interpolation: cubic
   discretize: # doesn't exist
      method: rouwenhorst
      mu: 3
```
This raises several potential issues:
- Dict-of-Dict are not nice in Julia since `{}` syntax disappeared. Potential solution, accept structures [:a=>b, :k=>[:a=>[0,0]]]` and convert them on the fly.
- another possibility would be to have special keywords replace a predefined subtree of options. So `time_it(.., grid=some_dict)` would merge `some_dict` with the subdict `options:grid`.
- fields outside of `options` like `exogenous` or `domain` would need additional kwargs, which is a bit annoying. So the current behaviour implies that to set them one needs to modify model data and make a `set_calibration` call. If the second option is kept, it would be aanother special case of it.
- in the current python implementation types are given by a key `type` but it's `tag` in the julia version. `type` seems more natural to me in both cases as tag really is inherited from yaml vocabulary
- another question is how to deal with excess parameters: with the merging mechanism, one can end up with a structure `Cartesian(min=, max=, orders=, mu=3)`. At which point should the `mu` be dropped ? Should a warning be issued ?

### DIFF
--- a/src/algos/time_iteration.jl
+++ b/src/algos/time_iteration.jl
@@ -103,11 +103,11 @@ If the stochastic process for the model is not explicitly provided, the process 
 * `dr`: Solved decision rule.
 """
 function time_iteration(model, dprocess::AbstractDiscretizedProcess, init_dr;
-      verbose::Bool=true, maxit::Int=100, tol::Float64=1e-8, details::Bool=true)
+          verbose::Bool=true, maxit::Int=100, tol::Float64=1e-8, options=Dict(), details::Bool=true)
 
     # get grid for endogenous
-    grid = model.grid
-    # grid = CartesianGrid(gg.a, gg.b, gg.orders)  # temporary compatibility
+    grid = get_grid(model, options=options)
+
 
     endo_nodes = nodes(grid)
     N = size(endo_nodes, 1)
--- a/src/algos/time_iteration.jl
+++ b/src/algos/time_iteration.jl
@@ -102,11 +102,16 @@ If the stochastic process for the model is not explicitly provided, the process 
 # Returns
 * `dr`: Solved decision rule.
 """
-function time_iteration(model, dprocess::AbstractDiscretizedProcess, init_dr;
-          verbose::Bool=true, maxit::Int=100, tol::Float64=1e-8, options=Dict(), details::Bool=true)
+function time_iteration(model::Model, dprocess::AbstractDiscretizedProcess,
+                        grid, init_dr;
+          verbose::Bool=true, details::Bool=true,
+          maxit::Int=100, tol_η::Float64=1e-8,
+          solver=Dict())
 
-    # get grid for endogenous
-    grid = get_grid(model, options=options)
+    if get(solver,:type,nothing) == :direct
+        return time_iteration_direct(model, dprocess, grid, init_dr;
+                  verbose=verbose, details=details, maxit=maxit, tol_η=tol_η)
+    end
 
 
     endo_nodes = nodes(grid)
@@ -142,7 +147,6 @@ function time_iteration(model, dprocess::AbstractDiscretizedProcess, init_dr;
 
     # loop option
     init_res = residual(model, dprocess, endo_nodes, x0, p, dr)
-    it = 0
     err = maximum(abs, stack0(init_res))
     err_0 = err
 
@@ -150,7 +154,8 @@ function time_iteration(model, dprocess::AbstractDiscretizedProcess, init_dr;
     verbose && println(repeat("-", 35))
     verbose && @printf "%-6i%-12.2e%-12.2e%-5i\n" 0 err NaN 0
 
-    while it<maxit && err>tol
+    it = 0
+    while it<maxit && err>tol_η
 
         it += 1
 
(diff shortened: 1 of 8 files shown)


sglyon:
This makes things a little more mysterious than explicit handing, but this kind of operation sounds great for a macro.

I had in mind something like

```julia
@mergable_options function time_iteration(args...; kwargs...)

end
```

And that would alter `time_iteration` in two ways:

- Add one more keyword argument named `options` at the end of the `...`
- Merge the options passed to this argument with any potential  arguments inside the model. 

Thoughts?

sglyon:
Oh and one thought about resolving the dict of dict issue.

I think we could provide something similar to what Gadfly does. Here's an example from their docs:

```julia
plot(x=1:10, y=2.^rand(10),
     Scale.y_sqrt, Geom.point, Geom.smooth,
     Guide.xlabel("Stimulus"), Guide.ylabel("Response"), Guide.title("Dog Training"))
```

Notice a few things:

- keyword arguments can come before or after positional ones
- All "options" have a type where the constructor arguments provide the values for that option
- As few or as little `Scale` or `Geom` objects can be passed and things will just work.

For us we would need a Julia type for each option (I think we were planning on this anyway) and we would probably want to have a keyword-only constructor for each of them.

I could imagine constructing the options like this:

```julia
Options(SmolyakGrid(n=15), Discretize.rouwenhorst(mu=3)) 
```

albop:
@sglyon : with your two comments you open many possible directions. 
On the second one: are you suggesting to use this syntax to replace dicts-of-dicts only, or to enhance most of the functions ? I'm not a big fan of Gadfly's syntax, which never does simple things, but I still see the value of your second example, if there is a way to make the mapping kind of systematic.
On the merge macro, I don't totally understand what the behaviour would be.
One thing I didn't mention before is that we have in general three structures. I want the "optimization options", I can have those specified in the yaml file, the default ones, and those given at the last minute when calling the function.

sglyon:
The merge macro would simply be there because I'm lazy... we'd write it in front of all the routines we'd like to enable to accept runtime options and have them merge with options in the model object. The macro would then add the keyword augment an do the few lines of code necessary to do the recursive merge. 

It is probably not worth it though as your current implementation just adds a keyword argument (not a big burden on us) and then calls `get_options` with the value of the kwarg, so that also isn't a big burden. If there were more repetitive lines of code required to enable option passing then we might think about the macro idea.

You are right about there being three sets of options. In my mind precedence should be runtime  > yaml > default. If you agree, then I think we can just do two recursive merges to start from default, then bring in yaml overrides with the first merge, then bring in runtime overrides with the second merge.

I'm happy to keep brainstorming on syntax. If you have more ideas for avoiding the dict of dicts please share. Perhaps the simplest alternative is just to have types for all options and make users build up the options struct using keyword arguments for type constructors

albop:
Yes, let's keep brainstorming. I suggest we defer the meta-option system until post 0.5.
Now, I'll try to make a list of important options for available routines, and that might be useful, to understand the option system.

albop:
Actually, we could try to list the common options to all algorithms and see whether that gives us some insights about how to make progress. I'll make a first pass.

src/algos/time_iteration.jl
```diff
@@ -102,12 +102,17 @@ If the stochastic process for the model is not explicitly provided, the process
 # Returns
 * `dr`: Solved decision rule.
 """
-function time_iteration(model, dprocess::AbstractDiscretizedProcess, init_dr;
-      verbose::Bool=true, maxit::Int=100, tol::Float64=1e-8, details::Bool=true)
+function time_iteration(model::Model, dprocess::AbstractDiscretizedProcess,
+                        grid, init_dr;
```
sglyon: It looks like we have added a few methods below to allow `grid` to be either a positional or keyword argument. 

Is this intentional?

sglyon: This comment also applies to the other solution algorithms like `time_iteration_direct`

src/algos/value_iteration.jl
```diff
@@ -168,16 +167,17 @@ Solve for the value function and associated decision rule using value function i
...
+    discount_symbol=:beta,
+    maxit::Int=1000, tol_x::Float64=1e-8, tol_v::Float64=1e-8,
+    optim_options=Dict(), eval_options=Dict(),
+    verbose::Bool=true, details::Bool=true)
 
-    # get grid for endogenous
-    grid = model.grid
 
-    β=model.calibration.flat[:beta]
+    β=model.calibration.flat[discount_symbol]
```
sglyon: I like this change!!!

src/algos/time_iteration.jl
albop: yes, it is intentional: the basic version of the function tries to take only "real" julia objects as arguments doesn't make any lookup in the symbolic data.

src/algos/value_iteration.jl
albop: wasn't expensive...

albop:
OK then I'll merge as is

PR #53, opened 2017-06-02
Sl/cleanups

sglyon:
Ok so I know this ways 35 commits -- but it really isn't that bad.

I needed to get tests passing so I can set up documentation. To do that I started looking through the code and got "distracted" by some things that I found that could be cleaned up.

So, I went through each file and 

- tidy up formatting
- Ensure we have no deprecation warnings on 0.6
- Fix some things to use "best practices"

Each file that I touched (All of them) have an associated commit or two with that type of work. They should all be prefixed with `CLN`

Then, I did make a handful of other changes

- Added docstrings a few places (commit msg header `DOC`)
- Fixed a few bugs (`BUG`)
- Use quantecon's routines for simulating Markov chains  https://github.com/EconForge/Dolo.jl/commit/b50fab7dde4c6693b63acaa6cc6a801a5719a81d
- Make data field of `Model` and `SModel` have type `Dict{Symbol,Any}` instead of `Dict`
- Added some methods to time_iteration_direct so the API of that routine matches the api of `time_iteration` https://github.com/EconForge/Dolo.jl/commit/85d4c9750cad26fa825003fead2bcffc837be106
- Simplified the API methods for `simulate(::Model, ...)` https://github.com/EconForge/Dolo.jl/commit/bbf4ece5448568e40859046fae82f4967fbd1527
- Fixed tests (last few commits)
- Updated REQUIRE

I'm planning on building on this very soon to get some autogenerated documentation up, so I'll probably merge quicker than I normally would


albop:
Thank you very much for this !

A few comments:
- for the discrete processes, I am not sure that simulating indices is all we need. Actually, it would feel natural to me if simulate(process) always returned the same type of data (i.e. continuous valued) no matter what the type of process is. Which would advocate for a simulate_indices method, instead of a simulate_values one.
- time_iteration_direct: I removed the convenience calls, with the goal of removing it in the medium run in favour of an option for time_iteration. It is exactly the same algorithm but equilibrium equations are solved directly, without a nonlinear solver. So exposing them as two different functions confuses users. There already is a an option `solver=:direct` which does that. In the short term it cannot do much harm though, we can just leave these functions undocumented.
- the `simulate(model::AModel)`: beware of these api changes as these methods are currently used in London. Also it is still not fully consistent. For instance there are issues over the ordering of s0 and m0. Switching both could have very unexpected results. Can we get that one as a separate PR ?

sglyon:
- RE simulating discrete process. So the original method simulated indices (this was actually implicit assumed in the model simulation routines) and had a bug for trying to return the values. I've just split them into two different routines right now. I'd be happy to use `simulate` and `simulate_indices` (like we do in quantecon), but the reason I kept it so that the simulate method returned indices is that it is used as the driving process in model simulation (see [here](https://github.com/EconForge/Dolo.jl/blob/0c3e25c20b86f3ca04abe606de4514bef115e7d6/src/algos/simulation.jl#L133) on master branch and [here](https://github.com/EconForge/Dolo.jl/blob/sl/cleanups/src/algos/simulation.jl#L146) in this PR). If we wanted to have `simulate` always return values instead of values for some processes and indices for Markov processes, then all we'd have to do is swap the names in this PR and call `simulate_indices` from the model simulation code. If you are ok with that I'm happy to make the change.
- re `time_iteration_direct` methods. I think that it is a good idea in the long run to have direct be an option for `time_iteration`, with the condition that docs are very clear that the option exists and what it does. I added back the convenience methods for now b/c we called them in the tests and it was the least painful way for me to get the algorithm tests running
- re `simulate(::Model)` routines. I can take these changes out of the PR. I think we should have a separate issue where we decide what api we really want as I don't really like what was there or what I have here.

albop:
- simulating discrete processes: the switch would indeed break some other methods. Since the current behaviour is currently unchanged let's do that as a separate issue.
- `simulate(model::AModel)`: let's discuss it separately. The current system might have a few issues but also has very cool features, like the ability to solve a model using a discretized Ar1 as a markov process but simulate it as a continuous one.

PR #55, opened 2017-06-10
ENH: update to use Dolang's new smart function routine

sglyon:
Should be merged with https://github.com/EconForge/Dolang/pull/31

### DIFF
--- a/src/algos/time_iteration.jl
+++ b/src/algos/time_iteration.jl
@@ -107,14 +107,13 @@ function time_iteration(model::Model, dprocess::AbstractDiscretizedProcess,
     end
 
     endo_nodes = nodes(grid)
-    N = size(endo_nodes, 1)
+    N = n_nodes(grid)
     n_s_endo = size(endo_nodes, 2)
     n_s_exo = n_nodes(dprocess)
 
     # initial guess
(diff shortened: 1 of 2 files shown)


albop:
The changes seem easy. I'll merge it as soon as I get a better notion of what it does.

sglyon:
Ok cool, once we get the failing test issue sorted out in Dolang, then this should be good to go.

the punchline of this change is that `make_method` no longer exists. Dolang now only has `make_function`, which will produce a generated function that can be evaluated at any order of derivative. 

Note that derivatives are only supported for functions with one vector argument with which to take the derivative (e.g. functions like `func(V, p)` -- where `p` is parameters)

albop:
I have just tested it today: it works with SymEngine.jl but not Calculus.jl which will pull many dependencies (including python).
 I'm still going to merge it as is. According to SymEngine github's it should be installable under windows. I"ll try to test it by next week-end. If it fails, we'll need a quick emergency plan.

sglyon:
Ok thanks for checking. 

Do you still have any output from when it failed under calculus? I'm really surprised by that given that we tried to be defensive about that case and Travis over at dolang passed without symengine

PR #56, opened 2017-06-12
Albop/prettify

albop:
This PR does 3 things:
- improve the output of time_iterations / value_iterations
- fix model printing in the notebook
- remove details options for VFI/TI

Note that value_iteration seems broken as the call to Optim doesn't work anymore.


sglyon:
Looks good to me.  Thanks

PR #61, opened 2017-06-17
FIX: changed call to Optim.minimize

albop:
Nothing controversial here. Don't really know why it would work but it does...


sglyon:
CI is broken here because we can't precompile Dolang anymore (see discussion here: https://github.com/JuliaLang/METADATA.jl/pull/9831)

We need to turn off precompilation on Dolo to get this working again.

albop:
Travis still failing with:
```
 Test threw an exception of type MethodError
  Expression: fc1 == fc2
  MethodError: no method matching start(::DataStructures.OrderedDict{Symbol,Float64})
  You may have intended to import Base.start
  Closest candidates are:
    start(!Matched::Dolo.TimeIterationLog; verbose) at /home/travis/.julia/v0.6/Dolo/src/algos/time_iteration.jl:85
```
This comes from the comparison between two OrderedDict in a test file. I tried to add `import  Base:start` and `import DataStructures:start` without success.  Any idea ?

albop:
Now another test failing. It is not failing on my laptop and I have no idea what's happening. @sglyon : could you have a look ?

albop:
Update: the tests pass with Symengine, not without.

sglyon:
That's odd. 

Is the error in Travis?

albop:
Yes but I get the same one if I uninstall symengine on my laptop

albop:
@sglyon : I'm going to merge that one, and open another Travis specific PR, so that I can continue the work on master.

PR #63, opened 2017-06-19
Improved Time Iteration

AnasZa:
ITI method added to Dolo Module. Works for all the models in the example file.


albop:
Thank you very much @AnasZa !  All tests pass, which is very unexpected. There must be something wrong ;-)
@sglyon : I have checked the algorithms and they are correct, while being a bit faster than regular time_iteration. Since it is still semi-experimental grade (as Dolo.jl), I'll merge it as is so people can play with it (and I can teach it on wednesday). Let's defer performance improvements and cleanup to later, and to afferent issues.

PR #64, opened 2017-06-20
MacroTools min version 0.3.3

tkelman:
looks like prewalk didn't exist before https://github.com/MikeInnes/MacroTools.jl/commit/17521b68054f1915d073fba30a1dc547ab912c3f


sglyon:
Thanks!

I forgot to remove the clone of Dolang after it was published the other day. 

I should also remove the clone of splines as that is not needed anymore

PR #65, opened 2017-06-20
Albop/perfect foresight

albop:
This is an initial implementation of the perfect foresight algorithm, with the following restrictions:
- it makes the assumption that complementarities are constant (this should actually be checked as a model feature)
- it doesn't use derivative informations, hence could be sped-up by many orders of magnitude
- the API is quite rudimentary.

### DIFF
--- a/src/algos/perfect_foresight.jl
+++ b/src/algos/perfect_foresight.jl
@@ -1,0 +1,117 @@
+
+function perfect_foresight(model, exo; T=200, verbose=true, complementarities=true)
+
+    T = T+1 # compat with t=0 conventions
+
+    T_e = size(exo, 1)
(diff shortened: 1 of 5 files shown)


src/algos/perfect_foresight.jl
```diff
@@ -0,0 +1,117 @@
+
+function perfect_foresight(model, exo; T=200, verbose=true, complementarities=true)
```
sglyon: It looks like we assume `exo` is a matrix of numbers. 

if that is correct, I'd like to put that in the signature as a form of documentation

sglyon: Great thanks! This is perfect for now

PR #71, opened 2017-06-21
RFC: simulate definitions routine

### LINKED ISSUE #70: IRFs for variables in "definitions"
It would be useful if Dolo.response or Dolo.simulate returns  an output also for the variables in the section "definitions".

sglyon:
This takes a stab at #70 

We should be able to call `evaluate_definitions(model, data, model.calibration[:parameters])` where `data` is the current output of either `simulate` or `tabulate`.

```

import Dolo
model = Dolo.yaml_import(joinpath(Dolo.pkg_path, "examples", "models", "sudden_stop_ar1.yaml"))
ti_res = Dolo.time_iteration(model);
data = Dolo.simulate(model, ti_res.dr, N=5, T=10)
defs = Dolo.evaluate_definitions(model, data, model.calibration[:parameters])
```

The underlying code comes with some flexibility in being able to specify arbitrary expressions to be evaluated on the simulated data (e.g. we aren't limited to `model.definitions`).

We could put this in the `simulate` and `tabulate` methods, but I didn't do it yet. What do you think @albop  and @AnasZa ?

Closes #70

### DIFF
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -27,7 +27,7 @@ import Distributions
 using AxisArrays
 
 # Compat across julia versions
-using Compat; import Compat: String, view
+using Compat; import Compat: String, view, @__dot__
 
 
 # exports
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -37,7 +37,8 @@ export arbitrage, transition, auxiliary, value, expectation,
 
        # mutating version of model functions
        arbitrage!, transition!, auxiliary!, value!, expectation!,
-       direct_response, controls_lb!, controls_ub!, arbitrage_2!, felicity!
+       direct_response, controls_lb!, controls_ub!, arbitrage_2!, felicity!,
+       evaluate_definitions
 
        # dolo functions
 export yaml_import, eval_with, evaluate, evaluate!, model_type, name, filename, id
@@ -82,10 +83,10 @@ include("numeric/processes.jl")
 include("numeric/decision_rules.jl")
 
 include("util.jl")
-include("symbolic.jl")
 include("calibration.jl")
 include("minilang.jl")
 include("model.jl")
+include("symbolic.jl")
 include("printing.jl")
 
 include("algos/steady_state.jl")
--- a/src/algos/perturbation.jl
+++ b/src/algos/perturbation.jl
@@ -18,8 +18,7 @@ end
 
 function Base.show(io::IO, pbr::PerturbationResult)
     @printf io "Perturbation Results\n"
-    @printf io " * Decision Rule type: %s\n" string(typeof(PerturbationResult))
-    @printf io "   * %s\n" show(pbr.solution)
+    @printf io " * Decision Rule type: %s\n" string(typeof(pbr))
     @printf io " * Blanchard-Kahn: %s\n" blanchard_kahn(pbr)
     @printf io "   * stable < %s\n" pbr.stable
     @printf io "   * determined < %s\n" pbr.determined
--- a/src/model.jl
+++ b/src/model.jl
@@ -211,6 +211,8 @@ type Model{ID} <: AModel{ID}
             eval(Dolo, code)
         end
 
+        eval(Dolo, build_definition_function(model))
+
         return model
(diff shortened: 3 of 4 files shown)


albop:
So what are the restrictions on the definitions ? Is it like Python ? (i.e. variables need to be contemporaneous)

sglyon:
Nope. The restriction is that the variable appears in the axis array. The assumption is that the column of the axis array contains period "0" obersevations of the variable. 

This routine will create a function that shifts times and pads with NaN as necessary.

albop:
Sounds great. Let me try it

albop:
- I don't get it: if I define `dy = y-y(-1)` or `t=y(1)-y` I get `symbol` undefined
- We should definitely add it by default to simulate/response/tabulate. Since it is compiled once for all, I don't see a good case for disabling it anyway.
- I wonder how much of it could be recycle to extend the  `eval_with` function to tackle expressions with timing like we did in Python a while ago (which is probably broken today). This is just an open thought, my prior being that it is another problem to solve.

sglyon:
Let me test it out some More on those examples.  I think the problem is that we accept shifted versions of variables, but not shifted versions of other definitions. 

Also, I think you already figured this out, but you'll need the latest commit of dolang for this to work

albop:
Actually, I wonder whether, independently from your function we are not making the working assumptions that definitions are contemporaneous. I remember considering it out of simplicity.

albop:
I just tested. Looks like we are. Which makes your pr instantaneously done, and another issue imminent...

sglyon:
I thought we assumed that we had to define them as time t variables, but that we can use them shifted in time. That's why the c(1)/c(0) in the Euler equation works. 

Is that what you are thinking too?

albop:
No, shifting works. But inside definitions there can't be any t-1 or t+1 neither on the lhs or the rhs.
This prevents us from defining `rho: c(1)/c` and use `rho` in the model for instance.

sglyon:
Oh ok. Makes sense. So everything has to be time t for definitions?

albop:
For now yes. So the definitions currently correspond exactly to `auxiliaries` (`y=a(m,s,x)`). In wonder whether we should expose it like that.
In the future it would be nice to allow different timings, but I know it creates a lot of complexities (like how we substitute defs), so it makes sense to do it in a second pass.

albop:
After a quick inspection of the code and running the tests locally, I'm merging that one with the following caveats:
- tests don't pass because Dolang was modified and the testsuite relies on latest published version ? (should we have two subcases with Dolang/master and Dolang/release ?
- We need to explicit somewhere (in the doc) that definitions are contemporaneous.

AnasZa:
Sorry for not participating in the discussion, but you did a very useful work ) Thank you! I am going to try to use it. Probably it would be nice to include it in the notebook with examples for users.

albop:
@anasza: maybe users don't need to use it directly, if you incorporate it
in the simulate/response/tabulate routines that you wrote.

AnasZa:
Oh, ok, so that is what I was not sure, I thought it is created to be used apart. So shall I put it as an option to the simulate/response/tabulate routines?  Or maybe even better to make it as a default (I mean without an option at all).

albop:
exactly

AnasZa:
I tried `Dolo.evaluate_definitions` on rbc_dtcc_ar1 model, and it works very nicely, however I have a problem with another model which is very close in spirits. I don't really understand the error message:  Please let me know if you have an idea what is going wrong @albop and @sglyon .

AnasZa:
The yaml file is actually in the dolo_models repository, named LAMP_4_julia.yaml.

sglyon:
Could you also please post the code you used when you called `evaluate_definitions`?

AnasZa:
`filename = "LAMP_4_julia.yaml"
model = Dolo.yaml_import(filename)
@time dr  = Dolo.improved_time_iteration(model)
sim2 = Dolo.response(model, dr.dr, :e_z)
Dolo.evaluate_definitions(model, sim2, model.calibration[:parameters])
`

PR #72, opened 2017-06-23
WIP: playing with product processes.

albop:
The basic example works but this is not finished.
One can now do: 
```
exogenous: !Product
    p1: !VAR1
         rho: 0.8
         Sigma: [[0.001]]
    p2: !MarkovChain
        values: [[0.0],[0.1]]
        transitions: [[0.97, 0.03], [0.1, 0.9]]
```
A few remarks:
- This works with two processes but one should be able to do it recursively. 
- All components are discretized as markov chains. In the future we could play with other combinations.
- I would have preferred to use a yaml list with possibly more than two elements but it is not possible with current yaml parsing.

### DIFF
--- a/examples/models/rbc_catastrophe.yaml
+++ b/examples/models/rbc_catastrophe.yaml
@@ -4,7 +4,7 @@ model_type: dtcc
 
 symbols:
 
-   exogenous: [z, u]
+   exogenous: [z, xi]
    states: [k]
    controls: [n, i]
    expectations: [m]
@@ -13,7 +13,7 @@ symbols:
    rewards: [u]
 
 definitions:
-    y: exp(z-u)*k^alpha*n^(1-alpha)
+    y: exp(z-xi)*k^alpha*n^(1-alpha)
     c: y - i
     rk: alpha*y/k
     w: (1-alpha)*y/n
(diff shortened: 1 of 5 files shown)


sglyon:
Cool!

Instead of `p1` and `p2` should we put the names of the exogenous symbols?

Is the yaml parsing  issue coming from YAML.jl?

albop:
Exogenous symbols wouldn't be enough as some subcomponents might be defined
for two joint variables. (Eg bivariate normal law)
I'm ok with requiring the user to pay attention to ordering for now.
Question 2: not quite sure. Maybe the way objects are constructed from the
yaml emitter (defined in Dolo). I think it assumes objects are constructed
using keywords.

examples/models/rbc_catastrophe.yaml
```diff
@@ -0,0 +1,90 @@
+name: Real Business Cycle
+
+model_type: dtcc
+
+symbols:
+
+   exogenous: [z, u]
```
sglyon: I think this is the same symbol (`u` ) as in the rewards section. We should fix this

albop: I fixed it. That's typically something a linter should be able to catch...

PR #74, opened 2017-06-26
Albop/fix travis

albop:
I am not sure travis is picking up the change. Might be an error in the travis.yml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,8 @@ matrix:
     - julia: nightly
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


albop:
Improved TCI time

sglyon:
wow, coverage really slows us down!

albop:
Yes ! Actually do we need it at all ?

sglyon:
no, definitely not right now. Maybe someday when we have better testing infrastructure

PR #76, opened 2017-06-27
Azh/response shocksize

AnasZa:
Allowing for changing the size of the shock in Dolo.response function + description of the response function.


albop:
Looks very good to me. Two small comments:
- some the doc looks outdated: like this reference to `sig_z`
- can you add some calls to new new variants in `test_algos.jl` ?

albop:
Todo bien ! I'll merge when Travis is done

PR #79, opened 2017-06-29
RFC: Smolyak from basis matrices

sglyon:
NOT READY TO MERGE

Here I leverage BasisMatrices.jl to implement Smolyak grids and Smolyak interpolation for models with more than one state and iid exogenous shocks. 

It's very likely that I'm just too tired to see how to do the other options, but I couldn't quite tell how to make smolyak work when the exogenous process had some sort of grid. Ideas welcome


albop:
@sglyon: I started to look at these changes. It's very, very useful but it changes a few things in decision rules that I am not sure I understand/agree with. I need to look more in detail. So let's discuss and not merge it right now, although I can't wait to have all of that.

A few comments:

0/ I know you had some doubts about the defining a grid but the examples seem very natural to me:
```
     @time Dolo.time_iteration(model, maxit=20, grid=Dict(:tag => :Random, :N => 200))
     @time Dolo.time_iteration(model, maxit=20, grid=Dict(:tag => :Smolyak, :mu => 3))
```
0.a. since a while ago, we import the yaml as a dict-of-dict structure with a `:tag` keyword everywhere. Could we make it a `:type` like in python ? `time_iteration(model, grid=Dict(:type => :Smolyak, :mu => 3))` would seem much more natural to me.

1/ Regarding your question. If exogenous grid is empty your interpolate only on the endogenous grid.  If exogenous is unstructured it's easy too: the d.r. is essentially a list of d.r. for each value of exogenous shock. If it's smolyak it's also easy (but you have to define the smolyak product of smolyak grids first). If it's cartesian it is conceivable (cartesian times smolyak) but probably more complicated.
1a. So the remaining question is how to get an exogenous process with a Smolyak grid instead of a Cartesian Grid. It sounds easy to do for a regular GDP discretization (like the one we do for VAR1). I'll have a look at that with @AnasZa .

2/ Adjustment to the definition of decision rules:
2.a. I don't see what you gain by redefining `DecisionRule{S<:Grid,T<:Grid,TCoef} ` (adding the Tcoef I mean).  In all subtypes, you actually set what TCoeff is. TCoeff doesn't seem to belong there.
2.b You define `CubicSplineDR{S<:Grid,T<:Grid} = DecisionRule{S,T,<:Vector{<:Array{Float64}}}`. The alternative we discussed a while ago was `DecisionRule{S,T,I}` where `I` is the interpolation type. I kind of like the simplicity of DR{Cartesian, Cartesian, Multilinear}. Is there any benefit in a subtype instead of a parameter ?
2.c. I I am not sure about the two following lines and what they mean:
```
+@compat const SmolyakDR{S<:Grid} = DecisionRule{S,SmolyakGrid,Vector{Matrix{Float64}}}
+@compat const CompletePolyDR{S<:Grid} = DecisionRule{S,RandomGrid,Vector{Matrix{Float64}}}
```
They seem to imply interpolation is only for IID, but that is wrong as it also make sense for Unstructured and Smolyak as the first grid. (see 1/)

3/ superficially cosmetic: in the definition of smolyak, it seems some of the fields are slightly too long
- B_nodes: B ?
- smol_params: parameters ?

sglyon:
Hey @albop thanks for the comments! I 100% agree that we shouldn't merge this as is. I was experimenting and decided to put the experiments up for others to take a look. As I was writing the code, it didn't all feel natural/optimal so I think we have some iterating to do before we are ready to merge this stuff.

I'm about to head out for a family activity day for the 4th of July, but I'll try to provide a quick response to some of your points.

(0) Defining the grids does indeed feel natural. I still think it is not fully general as I can't specify both the type of grid and the type of interpolation. Right now there is a single default interpolation type  for each combination of grid types. There's no reason why we can't have stuff like complete Legendre polynomials to use with the random grid instead of just compete monomials. (I've actually had some WIP code in BasisMatrices that implements complete Chebyshev, Legendre, Laguerre, and Hermite polynomials. It just needs to be tidied up).

(0.a) I'd vote to not change `:tag` => `:type` in YAML.jl. With `type` being  a Julia keyword and `tag` being the appropriate term in YAML nomenclature, I think tag is better for general use. That being said, I'm totally open to us using `type` in Dolo. For now we could just do a recursive pass through the keys of the dict of dict and make the change.

(1) Thanks for the explanation. I was able to figure out how to do the Smolyak and complete polynomials DRs when exogenous was both empty and unstructured. I implemented what you suggested above. I haven't, however, written code to combine an exogenous `CartesianGrid` with an endogenous `RandomGrid` or `SmolyakGrid`. I'm interested to see what you and @AnasZa come up with.

(2a and 2b) You are right that I always specify exactly what the third type parameter is. We could easily have three different structs that have strongly typed fields instead of doing what I did. The reason I chose the was just to save myself a little bit of extra work in the algorithms (e.g. time_iteration) and `CachedDecisionRule` always construct an instance of a single type `DecisionRule`, where the type of the third parameter would be based on the types of the two grids. I agree that it would be more natural to have the third type parameter be the type of interpolation. I didn't choose this because we don't have the `splines.jl` code wrapped up into a convenient type yet and I didn't want to do that while working on my quick prototype here. 

(2c) I'm not sure what you mean here. The type aliases for `SmolyakDR ` and `CompletePolyDR` only restrict the grid over endogenous states is `SmolyakGrid` or `RandomGrid`, respectively. The exogenous grid is allowed to be any subtype of `Grid`, which includes `UnstructuredGrid`, `EmptyGrid`, or `CartesianGrid` (maybe more?? I'd have to look). Although the type alias is flexible enough to support `CarteisanGrid`, I haven't implemented the code that does it yet (See my response to (1)). 

(3) I'm happy to make those changes to the field names on `SmolyakGrid`.

albop:
Hey @sglyon. Happy national day !
0/ If we call Dolo.time_iteration(model, maxit=20, grid=Dict(:tag => :Random, :N => 200), interpolation=:complete) or smthg like it, that still feel natural to me.
0.a/ I agree (`:tag` in `YAML`, `:type` in `Dolo`).
1/ Cartesian times Smolyak: I was saying "why not" on a purely theoretical level, with no intent of actually doing it...
2/a. How would you need the splines.jl code to be wrapped up ? I don't understand well the problem.
2/b. OK

sglyon:
Thanks!

0/ I would be fully happy if we came up with a solution that allowed the grid and interpolation type to be specified independently, but potentially depend on one another -- meaning the choice of interpolation could impact the specifics of the grid. 

1/ got it -- I think it sounds interesting and would like to see it, but don't have the time to do it myself (as you might have guessed)

2/a. The thought is that if we had one struct or type for each type of interpolation we supported, that third type parameter on decision rule could be the type of interpolation object we use. I think each of these types would need to have three basic functions:

1. A constructor that takes vectors of function values (e.g. vector of matrix) and stores coefficients in a struct 
2. Be able to update coefficients (`set_value!`)
3. Be able to be evaluated either through overloading getindex or call (my preference is for call)

Specifically for the splines.jl code I think we could make a `CUMultiSpline` that looks like this:

```julia
abstract type AbstractInterpolation end
# NOTE: we could get rid of the `N`, but then the actual coefficients would be weakly typed
# and evaluation might be slower than necessary...
struct CUMultiSpline{N} <: AbstractInterpolation
    coefs::Vector{Array{Float64,N}}
end
```


If we went this route and made all subtypes of `AbstractInterpolation` adhere to the proposed interface above then when we implement `DecisionRule` we could just hand off to the interpolation object for `set_values!` and evaluation.

Things still aren't 100% clear in my mind, but that is the type of thing I have in my head. Any thoughts/suggestions/criticisms are very welcome!

albop:
On 2a/: (decision rules and interpolation object).
I don't totally what an additional interpolation object adds to the mix. The methods you name are exactly the same as those for a DecisionRule object. DR is basically an interpolation object (in a general sense where we include perturbations). The only difference is DR can be defined on a product of two grids, while the interpolation object would be defined only on one.
The only use-case where it makes a difference to me is when defining (subtypes of) `DecisionRule{UnstructuredGrid, T}`. To interpolate/evaluate this kind of d.r., the only thing we need to know how to do is how to interpolate on a grid of type T. 
But for all other cases, the type of the first grid makes a difference. For instance, if you interpolate using cubic splines a `DecisionRule{Cartesian, Cartesian}` using BasisMatrices, you are going to precompute the basis functions corresponding to the exogenous grid  (i.e. do `B(m)B(s)c` with `B(m)` precomputed). In other words, you cannot reuse an interpolation method made for one grid.
Granted, it might be a quick way of getting one given interpolation routines to all grid topologies, but I don't think it should be the central scenario.

4. (new point)

This is not related to you PR only, but I'm thinking it would be great to be able to use the solution routines for DecisionRules/interpolation that are not part of Dolo source code. Something like:

```time_iteration(model, ..., ; DecisionRuleConstructor)```

There would be at least two reasons to do that:
1/ We could consider improvements like this PR. and make coexist with the existing d.r. objects. So we wouldn't have to commit to a new system too early, i.e. before measuring its impact in terms of API, performances...
2/ Playing with interpolation types that are either too new or not trusted as robust defaults, shouldn't require to make a fork of Dolo.
Make no mistake, I do want your smolyak and co. available as possible, but this PR is a good time to discuss this kind of changes.

**Comment by sglyon** [comment]
(2a) Good points. That the set of methods I described is the same as the current set of methods for `DecisionRule` was intentional. The thought was to implement the filtering and evaluation routines at the `AbstractInterpolation` level and then the implementation of those methods for the `DecisionRule` would just hand off to the underlying interpolation object. But, your points about the DR being defined over two grids instead of one is well taken and makes my proposal not work very well.

My proposal was an attempt at getting closer to what you mentioned above when you said

> The alternative we discussed a while ago was DecisionRule{S,T,I} where I is the interpolation type. I kind of like the simplicity of DR{Cartesian, Cartesian, Multilinear}. Is there any benefit in a subtype instead of a parameter

But, I don't think what I suggested is the right way to get there.

(4) I think this is a great idea. One way to do it would could be something like this:

```
function time_iteration{T<:AbstractDecisionRule}(...,  ::Type{T}, ...)

    # do stuff

    dr = T(args...)

    # use dr to solve model
end
```

Where we pass the `DecisionRuleConstructor` by just passing the type of a subtype of `AbstractDecisionRule`.

This would allow anyone to define their own subtype of `AbstractDecisionRule` outside of Dolo.jl.

Also, by us restricting the API to subtypes of `AbstractDecisionRule` signals that there is some contractual set of functionality that needs to be implemented by the user's type and isn't just a free-for-all with arbitrary functions

- - - 

> Make no mistake, I do want your smolyak and co. available as possible, but this PR is a good time to discuss this kind of changes.

I 100% agree -- we shouldn't be hasty here. I implemented this stuff as an experiment for me (us) to see _how_ these function approximation methods might fit in. I think the RFC period is still definitely open!

sglyon:
Ok I have rebased this PR off the current master (as of the morning of August 18th)

I have also done the following:

- Added type parameters to all grids specifying the dimensionality of the grid
- Added a type parameter `nx` to the `DecisionRule` type that specifies the number of controls approximated by the decision rule.
- Separated the decision rule code into separate files in a new a directory src/numeric/decision_rules as I found myself getting confused at exactly which parts of the former decision_rules.jl were abstract, which were for splines/smolyak/complete polynomials. Now there is a core.jl file in that directory containing the generic code and separate files for each specific type of DecisionRule. I'd be willing to go back on this if there is disagreement about whether or not this is a good change. 

Note that the code that adds dimensionality type parameters to the grids and DecisionRules is pretty ugly. No denying that. However, I'm hopeful/confident that after we finish the work with incorporating StaticArrays that the code in this PR can be cleaned up quite a bit. The main ugly part that I don't really like is that to get the type parameter I am computing the size of one dimension of an array and shoving that as the type parameter of a `Val` and picking that up via dispatch in a helper constructor. When we use StaticArrays the dimension of the arrays will already be known as a function type parameter so the Val business should be un-necessary.

sglyon:
One other thing to note is that this code doesn't support Julia 0.5. 

I could sprinkle in `@compat` in a couple dozen places if we really need 0.5 support. Given that JuliaPro is now released for Julia 0.6, I'd prefer not to do this, but I am not familiar with the constraints of other Dolo users.

robertdkirkby:
May or may not be relevant but I get impression you have used standard procedure of nested sets to create the Smolyak grids. (I can't be bothered reading the code to find out as you will know whether I am correct or misguided.) For speed you might want to consider the disjoint sets procedure of Judd, Maliar, Maliar & Valero (2014); according to them it is a fair bit faster. Both give exact same output in the end but the disjoint sets avoids creating some unnecessary objects that are created and then thrown away as part of the nested sets approach.

sglyon:
Hey @robertdkirkby thanks for the comment here.

We actually do use the disjoint algorithm by Judd et. al. -- with a few of our own improvements. When I wrote the Smolyak code (a few years ago) it was significantly faster (between 10-100x if I remember correctly -- @cc7768 might have a better memory) than the Matlab they provided with the paper.

robertdkirkby:
Sounds like you are not one but two steps ahead of me :)

albop:
@sglyon : I'm still looking at the PR. Meanwhile I've done some changes to the internal splines library, which does require StaticArrays (upcoming PR). So I propose the following two steps:
- merge the revamped splines into master
- do a minor release (soon) with no significant new features except bug fixes and a few performance improvements
- see whether your branch is still up to date, and if you want to include StaticArrays dependence. Then merge it into master.
- do a major release with the new interpolations and the linter. This one would depend on Julia 0.6 and come out somewhere in september.

sglyon:
That sounds like a good plan to me

This PR will be better/cleaner once I can use StaticArrays.

albop:
@sglyon : regarding 0.5/0.6, I'm now inclined to drop 0.5 before the next major release (i.e. 0.2) and break compatibility in master. In the unlikely case, we need to to an 0.1.x release for old julia pro users, we can try to quickly restore compatibility.

sglyon:
Ok sounds good to me1

albop:
I printed this PR and brought it for lunch.
Actually, in the way it is, it is a strict improvement as it allows for different grid types with matching interpolation, and no side effect that I can think of. So I'll merge it as soon as possible (@sglyon, I can rebase on master if you don't have time).
Some comments:
- the dimensionality code doesn't bother me that much (except I tend to use `N` for number of points and `d` for dimension). And the grid.jl can be revamped afterwards.
- the object design of the decision rules works for now, as long as we don't have many interpolation options for a given type of grid. For instance, `CubicSplineDR{S<:Grid,T<:Grid,nx} = DecisionRule{S,T,nx,<:Vector{<:Array{Float64}}}` could also correspond to the definition of multilinear splines. So it doesn't really work until you start creating wrapper objects for the last parameter IntType. However, it is reasonably close to the current design, so my vote goes for keeping it as is, and revisit later, when we want to provide interpolation options.*

* Actually, why don't we just have `LinearDR{S,T,nx}` and `CubicDR{S,T,nx}` be concrete types of `AbstractDecisionRule{S,T,nx}` and define a pseudo constructor `DecisionRule(S,T,nx,:linear)` or  `DecisionRule(S,T,nx,:cubic)` which returns the right object ? This could last until the Julia type system allows for more paremeterization options.

albop:
To be clear `DecisionRule(S,T,nx,:cubic)` could be `DecisionRule(S,T,nx,Cubic())` in case there are type inference issues.

sglyon:
Thanks again for the review.

Glad you think we have an improvement here. 

Once #109 is merged we should refactor this PR to build on top of that.

Once both are merged there is another round of cleanups/refactoring to do , but I think we are rapidly converging on a stable API/design.

sglyon:
I've screwed this up. Closing and will try to fix somehow...

PR #81, opened 2017-07-04
Azh/response shocksize

AnasZa:
Putting in order Dolo.simulate.jl + adding documentation.

### DIFF
--- a/src/algos/simulation.jl
+++ b/src/algos/simulation.jl
@@ -133,6 +133,26 @@ end
 ## methods which simulate the process
 ##
 
+"""
+This function simulates a model given a decision rule.
+
+# Arguments
+* `model::NumericModel`: Model object that describes the current model environment.
+* `dr`: Solved decision rule.
+* optional:
+  * `driving process` simulated series of a model's exogenous process.
+  If  `driving process` is not provided, then:
+  * `dprocess`: exogenous processes; default: model.exogenous.
+  * `s0::ListOfPoints`: List of initial state variable values; default: model.calibration[:states]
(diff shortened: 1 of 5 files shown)


src/algos/simulation.jl
```diff
@@ -122,62 +123,85 @@ end
...
-                  driving_process::AbstractArray{Float64,3})
-
-    s0 = model.calibration[:states]
-    return simulate(model, dr, s0, driving_process)
-end
+# function simulate(model::AbstractModel, dr::AbstractDecisionRule,
+#                   driving_process::AbstractArray{Float64,3})
+#     s0 = model.calibration[:states]
+#     return simulate(model, dr, s0, driving_process)
+# end
```
AnasZa: Ok ), I will delete.

src/numeric/processes.jl
```diff
@@ -385,6 +391,8 @@ function AgingProcess(mu::Float64, K::Int)
         transitions[i,1] = mu
     end
     transitions[end,1] = 1
+    dp = DiscreteMarkovProcess(transitions, values)
+    dp.i0 = 2
     return DiscreteMarkovProcess(transitions, values)
```
AnasZa: Yes, I am correcting.

sglyon:
I'm definitely in favor of keyword argument functions that allow us to delete the many methods we have now with positional arguments. 

I think it is very natural to have the initial index as part of the definition of the discrete Markov process. In fact , the definition of Markov chain in chapter two of Ljungqvist and Sargent includes "an initial probability distribution over states" -- we've chosen a degenerate distribution here!

Issue #33, opened 2017-03-22
Covariance matrix in the yaml file


albop:
@AnasZa found an annoying bug with the way covariance matrices are dealt with: when specifying the covariance matrix of a normal shock in a yaml file, the current behaviour of Dolo.jl is currently inherited from Distributions.MvNormal:

- `sigma: 0.1` translates into the covariance matrix `[[0.01]]`
- `sigma: [0.1, 0.1]` translates in the covariance matrix `[[0.01,0], [0,0.01]`
- `sigma: [[0.01,0], [0,0.01]` is the covariance matrix

Now this is inconsistent with the way yaml files are written (and how dolo.py parses them):

- `sigma: 0.1` is the same as `sigma: [0.1]` is the same as `sigma: [[0.1]]`

I see several ways to resolve the problem:
1/ broadcast the value in the yaml file to a 2d matrix (i.e. choose the python behaviour)
2/ choose and explain the Distributions behaviour and possible write `sig` instead of `sigma` so that it is *less* clear whether it's a vector of a matrix... (as well as make the notation consistent with Distributions).
3/ impose that users always write a matrix in the yaml file, raise an error otherwise
4/ introduce different argument names expecting different types of arguments: `sigma: ` would accept scalar standard deviation, `Sigma` would accept  a matrix only.

Option 1/ and 2/ are both feasible. Which one is best depends on what real people would naturally expect. I would tend to exclude the status quo 1/, but am not too sure about 2/.
Option 3/ is perfectly safe and totally unexciting. Also it doesn't break any existing yaml file.
Option 4/ reintroduces additional flexibility, but in a less canonical way. 

What do other think ?

sglyon:
So, if I understand correctly, we are inconsistent in how we interpret the value of the `sigma` field? Sometimes we treat it as a standard deviation, other times we treat is a vector of standard deviations terms (for independent variables), and other times it is a full covariance matrix?

albop:
Yes, if the user enters a covariance we are consistent. But if he/she enters a scalar or a vector, the way we convert it to a matrix is inconsistent.

sglyon:
Should we just always assume they enter a variance?

So if it is a scalar it is the variance of a 1d Gaussian.

If it is a vector it is the diagonal of a _covariance_ matrix.

albop:
That is exactly what we currently do on the python side (and in existing modfiles). But it's not the convention in Distributions.jl (see above)

sglyon:
[...]

PR #85, opened 2017-07-07
Adding an error message in grid.jl. Fixing a bug in discretize(mvn::MvNormal).

AnasZa:
About the bug: 
A model with 2 iid shocks. The yaml file excepts Sigma of exogenous process as 2-element Array{Array{Int64,1},1}:
```
exogenous: !Normal
       Sigma: [[sig_a^2], [sig_b^2]]
```

But discretizing model.exogenous it raises an error, because: `ishermitian(A) || non_hermitian_error("chol")`. I fixed it, however I wonder whether it is more logical to change the yaml file to accept: 
```
exogenous: !Normal
       Sigma: [[sig_a^2 0]; [0 sig_b^2]]
```

### DIFF
--- a/src/numeric/processes.jl
+++ b/src/numeric/processes.jl
@@ -90,7 +90,7 @@ MvNormal(sigma::Float64) = MvNormal(reshape([sigma], 1, 1))
 
 function discretize(mvn::MvNormal)
     n = fill(5, size(mvn.mu))
-    x, w = QE.qnwnorm(n, mvn.mu, mvn.Sigma)
+    x, w = QE.qnwnorm(n, mvn.mu, diagm(vec(mvn.Sigma)))
     DiscretizedIIDProcess(x, w)
 end
 
@@ -211,7 +211,7 @@ function discretize(::Type{DiscreteMarkovProcess}, var::VAR1; N::Int=3)
 
     # it would be good to have a special type of VAR1 process
     # which has a scalar autoregressive term
-    R = var.R
+    R =  var.R
     d = size(R, 1)
     ρ = R[1, 1]
     @assert maximum(abs, R-eye(d)*ρ)<1e-16
--- a/src/model.jl
+++ b/src/model.jl
@@ -125,6 +125,10 @@ function get_calibration(model::ASModel)
     end
     # so far _calib is a symbolic calibration
     calibration = solve_triangular_system(_calib)
+    # if  isempty(calibration)
+    #   msg = string("bla")
+    #   error(msg)
+    # end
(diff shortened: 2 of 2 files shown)


src/numeric/processes.jl
```diff
@@ -90,7 +90,7 @@ MvNormal(sigma::Float64) = MvNormal(reshape([sigma], 1, 1))
 
 function discretize(mvn::MvNormal)
     n = fill(5, size(mvn.mu))
-    x, w = QE.qnwnorm(n, mvn.mu, mvn.Sigma)
+    x, w = QE.qnwnorm(n, mvn.mu, diagm(vec(mvn.Sigma)))
```
sglyon: This change doesn't look correct to me. Isn't `mvn.Sigma` already the covariance matrix?

AnasZa: Nope. In this case it was just a 2×1 Array{Float64,2}.

sglyon: Which case was this? Do you have an example where this came up?

If we did get a 2x1 array then it sounds to me like the issue is when we construct the instance of MvNormal

AnasZa: Probably you are right. However, it seems like it is matter of convention.. or not. Please have a look at the example I am sending you by email.

albop: I agree that here Sigma should always be a square matrix.

albop: and also agree that we should require the yaml file to have a matrix as in
```
exogenous: !Normal
       Sigma: [[sig_a^2 0]; [0 sig_b^2]]
```
and not promote automatically other forms like `Sigma: sig_a^2` or `Sigma: [sig_a^2, sig_b^2]`. They look like they can be intepreted easily (as a 1x1 matrix or a diagonal) but for now they create uncertainty (see bug #33).

AnasZa: So actually as it is not, yaml file accepts a Vector: 
```
exogenous: !Normal
       Sigma: [[sig_a^2 0], [0 sig_b^2]]```
 I don't know why it didn't work for me before. I will close this pull request.

Comment (albop) [inline src/numeric/processes.jl]: In your example above you had:
```
exogenous: !Normal
       Sigma: [[sig_a^2 0]; [0 sig_b^2]]
```
and I don't know what `;` means in a waml file. Maybe that was the issue ?

AnasZa: Sorry, it was a proposal how it should be, but not implemented. Or probably I haven't understood your remark.

PR #86, opened 2017-07-07
ENH: enforce square Sigma in !Normal exog spec

sglyon:
I think this should help with #85, but I would like a review just to be sure this is the behavior we want.

### DIFF
--- a/src/minilang.jl
+++ b/src/minilang.jl
@@ -91,6 +91,11 @@ function _build_exogenous_entry(data::Associative, calib::ModelCalibration)
         # need to extract rho an dSigma
         Sigma = eval_with(calib, data[:Sigma])
         Sigma = to_matrix(Sigma)
+        if size(Sigma, 1) != size(Sigma, 2)
+            sz = size(Sigma)
+            msg = "Covariance matrix must be square. Found a $sz matrix"
+            throw(DimensionMismatch(msg))
+        end
         return Normal(Sigma)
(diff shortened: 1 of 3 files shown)


sglyon:
Right now the current code will accept two things:

1. `sig` or `[sig]` will be promoted to `[[sig]]` for univariate Normal processes
2. Any other list of lists for more than one dimension **with the restriction** that the result is square. 

Is this acceptable? If we really want to be strict about disallowing scalars then I can implement that change, but it seems like something we will likely want to do eventually so the fact that we have it now seems like a bonus to me.

albop:
What happens with `[a,b]` then ?
My current thinking is you can mistake a scalar for a matrix, but you can't mistake a matrix for a scalar. So asking everything to be a matrix cannot lead to an error later, and we can make things more flexible in the future.
Problem with the scalar case is one is tempted to think of `Sigma: sig_z` as a standard devitation, not a 1x1 matrix. So until the thought has perished, my vote would be to ask the matrix to always be square, even if that's a bit painful.

sglyon:
`[a, b]` would throw an error.

I can make the change to only accept list of lists for now

albop:
I think it's fine as it is. We need to add somewhere in the doc that `Sigma` always contain variances or covariances.

sglyon:
Ok , sounds good.

Sorry I didn't get a chance to make that change -- I've been travelling.

PR #87, opened 2017-07-10
Azh/error msg

AnasZa:
Adding error messages. One when equation sub-sections are misspelled. Another one is about a matrix of grid points given in section: options-grid-orders.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -10,3 +10,4 @@ QuantEcon 0.10.1
 Distributions
 Compat 0.17.0
 Dolang
+StringDistances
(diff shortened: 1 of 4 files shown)


sglyon:
That's fun to use `StringDistances` to offer an informed guess at what the user wanted.

Make sure to add StringDistances to the `REQUIRE` file to it is installed with dolo.jl and on travis where our tests are being run.

AnasZa:
Thanks. I didn't know that.

albop:
That's one more step towards a real artificial intelligence...

PR #89, opened 2017-07-11
Edited perfect foresight section for algos.md

ezgioz:
Table seems nicer now.


albop:
This looks fairly good to me. Just saw two things in the text:

albop:
Thank you @ezgioz.
I just noticed two little things:
- initial state in the text is not consistent with the table (should be `\overline{s_0}`)
- in the arbitrage condition at the beginning of the paragraph, the expection brackets don't look nice. They should be square.

PR #95, opened 2017-07-13
DOC: Changes to exogenous process.

azheng25:


docs/src/modeling_language.md
```diff
@@ -281,89 +281,64 @@ issuing any warning.
...
-in the distribution section. A gaussian distrubution (only one supported
-so far), is specified by supplying the covariance matrix as a list of
-lists as in the following example.
+Here are examples of how to define different processes. Note the use of yaml tags.
+Normal Shock: It has mean zero and variance `Sigma`. 
+```{.sourceCode .yaml}
+exogenous:!Normal
+   Sigma: [[0.016^2]]
+```
+VAR1: `rho` is the persistence (only one allowed for now). `Sigma` is the covariance matrix. (Note: if a scalar `sigma` is given, it will be converted to `[[sigma]]` to indicate that it is the variance (not covariance) of the process).
```
sglyon: I think we should be clear that a scalar `sigma` is also not the standard deviation. That field will always represent a variance for a scalar process or a covariance matrix for a multivariate process.

albop:
@sglyon : I fully agree with your comment of course. @azheng25 is travelling so I'll merge this PR and propose a small change just afterwards.

PR #97, opened 2017-07-17
FIX: set_calibration was not working.

albop:
@sglyon, @azheng25 : this fixes two things:

- rename `set_calibration` to `set_calibration!` which seems to make more sense
- older, experimental, version used to change only the symbolic structure (in model.data["calibration"]). That was pretty useless. Now we update all calibrated values, (grid, domain, etc... not options).

Not that we still change one parameter at a time.

### DIFF
--- a/src/Dolo.jl
+++ b/src/Dolo.jl
@@ -42,7 +42,7 @@ export arbitrage, transition, auxiliary, value, expectation,
        evaluate_definitions
 
        # dolo functions
-export yaml_import, eval_with, evaluate, evaluate!, model_type, name, filename, id, features
+export yaml_import, eval_with, evaluate, evaluate!, model_type, name, filename, id, features, set_calibration!
 
 export time_iteration, improved_time_iteration, value_iteration, residuals,
         response, simulate, perfect_foresight, time_iteration_direct, find_deterministic_equilibrium, perturbate
--- a/src/model.jl
+++ b/src/model.jl
@@ -270,6 +270,22 @@ function set_calibration!(model::Model, key::Symbol, value::Union{Real,Expr, Sym
     model.calibration
 end
 
+function set_calibration!(model::Model, values)
+    for (key,value) in values
+        model.data[:calibration][key] = value
+    end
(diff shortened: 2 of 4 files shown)


sglyon:
Looks good. Tests would be best, but if you don't have time now we can come back around later

albop:
You guessed right...

albop:
@sglyon All right, revisiting, with a bit more good will I made two additions:
- a test file
- additional calls `set_calibration!(model, Dict(:beta=>0.99, :delta=0.09))` and `set_calibration!(model, beta=0.99, delta=0.08)`
BTW: I didn't manage to type the second arg as smthg like `Dict{Symbol, Union{Real, Expr, Symbol}}`. If you happen tell me, I'll make hte fix.

sglyon:
Nice!! Thank you for adding tests.

I also like the keyword argument version. The API seems very flexible now.

sglyon:
What was your question about the type of the Dict? I didn't totally follow

albop:
The versions with dictionary arguments are:
```function set_calibration!(model::Model, values)```
and 

``` 
function set_calibration!(model::Model; kwargs...)
    set_calibration!(model, kwargs)
end
```

Initially I wrote

**Comment by sglyon** [comment]
Ahh I see. I think the issue is that splatted keyword arguments are not thrown into a Dict by default. I think this method definition should work:

```julia
function set_calibration!(model::Model; kwargs...)
    set_calibration!(model, Dict(kwargs))
end

function set_calibration!{T}(model::Model, values::Associative{Symbol,T})
    for (key,value) in values
        model.data[:calibration][key] = value
    end
    model.calibration = get_calibration(model)
    model.exogenous = get_exogenous(model)
    model.domain = get_domain(model)
    # model.options = get_options(model) # we don't substitute calib here
    model.grid = get_grid(model)
    model.calibration
end
```


but I haven't tested that out so I'm only like 98% sure

albop:
This works, so I made the change.

sglyon:
Will merge when green

PR #101, opened 2017-07-25
Ezgioz/improve pf

ezgioz:
Instead of defining a matrix "exo" for shocks (order of shocks/size is important)

    # Provide the shocks with dictionary 
exo = Dict(:z =>  [0.00, 0.01, 0.02, 0.03, 0.04], :xi =>  [0, 0, 0, 0, 0, 0])

    # Not provided shocks will be equal to zero by default
    # Vector sizes for each shock can be different
    # The order the shocks are defined is irrelevant: pf fuction finds the index for each shock

### DIFF
--- a/src/algos/perfect_foresight.jl
+++ b/src/algos/perfect_foresight.jl
@@ -1,19 +1,25 @@
 """
 Document pf.
 """
-function perfect_foresight(model, exo::AbstractMatrix{Float64}; T=200, verbose=true, complementarities=true)
+function perfect_foresight(model, exo::Dict; T=200, verbose=true, complementarities=true)
 
     T = T+1 # compat with t=0 conventions
 
-    T_e = size(exo, 1)
     n_e = length(model.symbols[:exogenous])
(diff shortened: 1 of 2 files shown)


albop:
Thank you @ezgioz ! This is a welcome addition.
Here are a few comments (1/ is the main one):
1/ you should keep the older method of `perfect_foresight` which takes argument `exo::AbstractMatrix{Float64};`. What you need to is define a *new* method `perfect_foresight(..., exo::Dict{Symbol, Array}, ...`. This new method would construct the exogenous matrix of size `T_e` and pass it the the existing one.
2/ we don't have to assume that all shocks have the same have same length: `exo = Dict(:z => [0.00, 0.01, 0.02, 0.03, 0.04], :xi => [0, 0.1, 0])` should work. The method would construct a matrix with size T_e=5. (unrelated: backticks ` in markdown to quote simple pieces of code)
3/ you could send an error if `T_e>T`

PR #104, opened 2017-07-28
Ezgioz/improve pf

ezgioz:
Method defined as:
function perfect_foresight(model, exo::Dict{Symbol,Array{Float64,1}}; kwargs... )

Given a dictionary exo, constructs the matrix exo and calls the original method. Test file tests if all the results are the same.

- One shock
matrix vs dictionary 

- Two shocks
matrix vs dictionaries defined differently (different order for shocks or zero shock not explicitly provided)

Instead of  exo::Dict{Symbol,Array{Float64,1}}, tried a more general definition for dictionary type but 

> exo::Dict{Symbol,Array{Real,1}}
 or
> exo::Dict{Symbol,Any}

doesn't work.

So we expect Float64 as input. 

This may cause errors if user defines a zero shock as [0,0,0,0] for instance. Shall we define this argument as exo::Dict{} then?

### DIFF
--- a/src/algos/perfect_foresight.jl
+++ b/src/algos/perfect_foresight.jl
@@ -119,7 +119,7 @@ function perfect_foresight(model, exo::AbstractMatrix{Float64}; T=200, verbose=t
 end
 
 # Constructs the matrix exo given a dictionary exo and calls the original method
-function perfect_foresight(model, exo::Dict{Symbol,Array{Float64,1}}; kwargs... )
+function perfect_foresight(model, exo::Dict{Symbol,Array{Real,1}}; kwargs... )
 
   n_e = length(model.symbols[:exogenous])
   T_e = maximum(length(e) for e in values(exo))
--- a/src/algos/perfect_foresight.jl
+++ b/src/algos/perfect_foresight.jl
@@ -119,8 +119,8 @@ function perfect_foresight(model, exo::AbstractMatrix{Float64}; T=200, verbose=t
 end
 
 # Constructs the matrix exo given a dictionary exo and calls the original method
-function perfect_foresight(model, exo::Dict{Symbol,Array{Float64,1}}; kwargs... )
-
+function perfect_foresight(model, exo::Dict{}; kwargs... )
+  convert(Dict{Symbol,Array{Float64,1}}, exo)
   n_e = length(model.symbols[:exogenous])
   T_e = maximum(length(e) for e in values(exo))
   exo_new = zeros(T_e,n_e)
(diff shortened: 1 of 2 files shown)


albop:
Maybe you could take 'AbstractArray{Float64}' as input vectors and convert to floats ?

albop:
Sorry I meant 'AbstractArray{Real}'

ezgioz:
I wanted to define the new method being specific on key and more general for values. Something like:

> function perfect_foresight(model, exo::Dict{Symbol,Array{Real,1}; kwargs... )

Then 

> convert(Dict{Symbol,Array{Float64,1}}, exo)

to create the exo matrix as needed for the first method

However neither AbstractArray type nor one of below way to define the method works

> function perfect_foresight(model, exo::Dict{Symbol,Array{Real,1}; kwargs... )

> function perfect_foresight(model, exo::Dict{Symbol,Array{Any,1}; kwargs... )

> function perfect_foresight(model, exo::Dict{Symbol,Array{Any}; kwargs... )

I get errors (no method defined) in any case, both for Array{Int64,1} and Array{Float64,1} type of inputs. Which I don't understand, both should be sub-types of "Real"

So in the end I defined the method in a quite general way:

> function perfect_foresight(model, exo::Dict{}; kwargs... )

And convert the array accordingly on the function file before calling the first method:

> convert(Dict{Symbol,Array{Float64,1}},exo)

albop:
Maybe I gave you the wrong advice. Try `Union{AbstractVector{Float64}, AbstractVector{Int64}}`, instead of `AbstractVector{Real}`. That should work.

sglyon:
I think the best approach would be to try `AbstractVector{<:Real}`, which will do `Union{AbstractVector{T} \forall T <: Real}`

If we need to support 0.5 still then we just need to put `@compat` before that

albop:
Perfect ! Thx @sglyon

albop:
I made a quick test and I'm not sure your solution work:
```
julia> g(x::Dict{Symbol,AbstractVector{<:Real}}) = convert(Dict{Symbol, Vector{Float64}})
g (generic function with 1 method)

julia> g(Dict(:x=>[0.0, 0.9]))
ERROR: MethodError: no method matching g(::Dict{Symbol,Array{Float64,1}})
Closest candidates are:
  g(::Dict{Symbol,AbstractArray{#s1,1} where #s1<:Real}) at REPL[7]:1
```
I suggest we move the discussion about typing somewhere else. In the meantime, I'll merge as is.

PR #105, opened 2017-08-04
Azh/reordering simulate(n,n,t)

AnasZa:
Making simulation of var process faster. The code was simulating the process NxN times, instead of just N. Plus I reordered the input as (n,N,T).


albop:
Thank you !

PR #110, opened 2017-09-01
Sl/basis matrices

sglyon:
This really should be in #79 , but somehow things got funky... I had to do some git surgery and I think I confused github

Either way, I think this is ready to go


albop:
@sglyon : I have gone over all the changes there might be bugs (like smolyak in 1d or random with markov chains) but we could iron them out later on, potentially after we have switched data layout and written some tests for DR, that are independent of the models.
One question though: you still have definitions like: `const CubicSplineDR{S<:Grid,T<:Grid,n_x} = DecisionRule{S,T,n_x,<:Union{<:Array{<:Point},<:Array{<:Array{<:Point}}}}` not yet different concrete subtypes for different interpolation modes. Do you want to do that in this PR or in a different one, after I have merged it ?

albop:
ah, and travis is missing `sudo apt-get install libnlopt0`

albop:
Actually, let's merge now.

sglyon:
I was planning on waiting for a different PR to introduce concrete types for different interpolation methods.


Also if you think there are bugs (not missing functionality) please file an issue so we can sort that out.

Thanks!

PR #111, opened 2017-09-04
Fix: Added missing method

ezgioz:
> improved_time_iteration(model, init_dr);

was not working so added the method below

> function  improved_time_iteration(model, init_dr; grid=Dict(), kwargs...)
> ...


albop:
Looks good ! Thank you @ezgioz

PR #112, opened 2017-09-05
Concrete types for each interpolation flavor

sglyon:


albop:
Thanks @sglyon . One consequence of moving to `Point{d}` and `ListOfPoint{d}` is that we are no longer compatible with AxisArrays (as input to decision rules).
Indeed, the d.r. can't take AbstractArray as input anymore, as the memory layout of input is no longer indifferent. 
This is easy to fix, but I confess I hadn't paid that much attention to it.

albop:
I guess the best way to fix that is to replace all the calls `reintepret(Point{d}, s', (N,))` (which is very error prone if the `'` is forgotten) by a wrapper which makes a copy of s' when required.

sglyon:
Ahh that's right. Looks like the tests are failing there. I like your idea of the helper method to convert from matrix to list of points. It would be too easy to forget the transpose somewhere...

albop:
I reviewed the code and it seems OK, so I will merge without further delay. Thanks for adding the tests of decision rules.
Fixing this AxisArrays was useful as it made me realize there are many places where we use array views, which will need some changes.

PR #115, opened 2017-09-12
Azh/generator

AnasZa:
Adding a generator that would substitute 
```
for j in n_inodes(dprocess)
    M = dprocess.inode(i)
    w = dprocess.iweight(i)
    ...
end
```
by
```
for (w, M, j) in get_integration_nodes(dprocess)
    ....
end
```
This generator is supposed to take less memory, plus it excludes the future nodes which have probability 0 from an iterations loop.

### DIFF
--- a/src/algos/ITI_additional.jl
+++ b/src/algos/ITI_additional.jl
@@ -39,23 +39,24 @@ function euler_residuals(model, s::AbstractArray, x::Array{Array{Float64,2},1}, 
        m=(hcat([e' for e in m_prep]...))'
 
        for I_ms in 1:n_mst
-         for (w, M_prep, j) in get_integration_nodes(dprocess,i)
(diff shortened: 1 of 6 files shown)


AnasZa:
Hmmm, I rerun `Pkg.test("Dolo")` and I passed all tests.

albop:
Thank you @AnasZa : this is a welcome improvement. In particular, it should speed up the computation of the life-cycle models quite a bit.
As for Travis, the log says:
```
ERROR: LoadError: LoadError: LoadError: LoadError: syntax: incomplete: "function" at /home/travis/.julia/v0.6/Dolo/src/algos/ITI_additional.jl:4 requires end
```
Also, you can get proceed forward and remove the commented line. We still have them in the git history.
Lastly, for the record, we don't need to replicate the exogenous vector as in `M=(hcat([e' for e in M_prep]...))'` anymore. Since this is already fixed in another branch, you can leave it here.

PR #116, opened 2017-09-14
Azh/product gdp

AnasZa:
Adding a functions to discretize the product of  processes as GDP:
```
var=Dolo.VAR1(rho,Sigma)
var2=Dolo.VAR1(rho1,Sigma1)

pp=Dolo.ProductProcess(var,var2)

Dolo.discretize(Dolo.DiscretizedProcess,pp)
```
`Dolo.discretize(pp)` returns however `MarkovDiscretizedProcess`.


albop:
Thank you @AnasZa !
Apparently the new function `discretize(...)` is defined too early in the file `processes.jl` (Travis says ProductProcess not defined). You could move it to the end of the file, just after `discretize(::Type{DiscreteMarkovProcess}, pp::ProductProcess;)`. 
Actually, you can copy the definition of this method to pass arguments to the subprocesses.

albop:
As for the `ProductGrid` function, I suggest:

albop:
Also I suggest:
- to rename `ProductGrid` as `Product`, since the argument type made it clear what it does
- in the function `Product(gdp_1, gdp_2)`  you can remove the checks. If the grids are not of the right type, the ProductGrid inside will complain.

AnasZa:
Ok. But what about the fact that by default it is discretized as MC rather then GDP as in simple case?

albop:
It's fine. There has to be a default, right ? MC is a good one.

albop:
It seems to work. But there is no test for the new functionality. Could you add one ?

PR #121, opened 2017-10-11
FIX: condition in options - size of order edited

ezgioz:
length(orders)!=length(model.calibration[:states])


albop:
That works. Thank you !

Issue #125, opened 2017-12-09
Special object to store discretized decision rule.


albop:
Currently we store the controls on the grid in any exogenous state `i_m \in [0,N_m]` as a `Vector{ListOfPoints{d}}` where the size of the vector is `N_m` and the size of each `ListOfPoints` is `N`.

Given an element $x \in Vector{ListOfPoints{d}}$, we tediously perform the following operations:
1/ compute error `maxabs(x)` which is the supremum norm over all the elements
2/ access i-th list of points `x[i] isa ListOfPoints{d}`
3/ convert to/from a flat vector

Last item is problematic: we make a copy of the data in x which is a shame. The question I'm asking there is whether we could have an object, whose data would be one single contiguous vector that could also be treated as a Vector of ListOfPoints.

There is at least one problem I foresee:
- we currently use reinterpret a lot, which converts vectors into other vectors. If I'm not mistaken in >0.7, it would return an AbstractArray instead. So the natural approach would  be to define x[ì] as a basic view over the data.* But we (I) have recently made the assumption everywhere that we are dealing with contiguous vectors. We'll have to reconsider, probably re-AbstractVectorize everything...

Some features that would be nice for this object:
- not assume that all lists of points have the same length
- not make precise assumption about the type of element. For instance it could work with `Point{d}` and `Jacobian{d}` as well (a square matrix).

I guess it should be possible to make something simple and robust here, so let's think it hard !

* Something like:
```
struct StangeArray{T}
    data::Vector{T}
    views::Vector{ArrayView{T}}
end
```

PR #126, opened 2017-12-20
ENH: special object for decision rules.

albop:
@sglyon : here is an attempt to resolve issue #125 .
One can do:
```
drs = DRStore([rand(32) for i=1:3])
drs.data :: Tuple{Vector{Float64}}
drs.flat :: Vector{Float64}
maximum(abs, cat(1, drs.data...)- drs.flat) # -> 0
drs.flat[10] = 1.2
maximum(abs, cat(1, drs.data...)- drs.flat) # -> 0
```

The fields `drs.data`  can contain the decision rule as we store it right now (almost, it is a tuple instead of a vector so that it is immutable) and `drs.flat`  represents it in all states as one single vector. Trick is both fields refer to the same part of memory and are always synchronized. I believe there is no memory cleaning problem but that would require careful thinking.

I have hesitated as to how `drs` itself should behave. In the end, I implemented the behaviour from before (so `drs[1]` returns a vector) but didn't use it. The reason is to backward compatibility, but I ended up not using `drs` directly, using instead `drs.data` most of the time.
The alternative would `drs` to behave as `drs.flat`, at which case it could derive from AbstractVector. I can see how we would get some operations for free, but haven't found one very good reason. An idea would be to do `drs[i]`  for flat indexing and `drs[G(i)]` or smthg like that for group indexing.

So far I have only applied it to `time_iteration_direct` (memory 350->20) and would like some early feedback before generalizing it. Any thought ?


src/drstore.jl
```diff
@@ -0,0 +1,74 @@
...
+# Main.norm(ds::DRStore{T},k::Int) where T = norm( (norm(e,k) for e in ds.flat), k)
+distance(ds1::DRStore{T}, ds2::DRStore{T}) where T = maximum( k->norm(k[1]-k[2],Inf), zip(ds1.flat,ds2.flat) )
+# distance(ds1::DRStore{T}, ds2::DRStore{T}, k::Int) where T = norm( (norm(e[1]-e[2],k) for e in zip(ds1.flat,ds2.flat)), k )
+
+total(ds::DRStore{T}) where T = sum(length(e) for e in ds.data)
+
+function DRStore(::Type{T}, dims::AbstractVector{Int}) where T
+    L = sum(dims)
+    fD = zeros(T, L)
+    ptr = pointer(fD)
```
sglyon: Yikes 😨 ! I don't think I'v ever seen `pointer` and `unsafe_wrap` used out in in the pure Julia wild (e.g. not interfacing with C libs) 

Are we confident that this is safe and works correctly? I would guess that it does given your experiments, but it makes me a little scared.

What type of penalty would we get if we were to construct `D` using SubArrays (e.g. using `view`) instead of `unsafe_wrap` and pointer offsets?

src/drstore.jl
```diff
@@ -0,0 +1,74 @@
...
+end
+
+# this one reuses data in place
+# should it be called DRStore! ?
+function DRStore(fD::Vector{T},dims::AbstractVector{Int}) where T
+    L = sum(dims)
+    ptr = pointer(fD)
+    sz = sizeof(T)
+    offsets = [[0]; cumsum(dims)[1:end-1]]*sz
+    D = [unsafe_wrap(Array{T}, ptr+offsets[i], dims[i]) for i=1:length(dims)]
```
sglyon: I think we should probably be testing that `fD` is the  correct length, right?

`unsafe_wrap` will let us have elements of `D` defined over memory that extends beyond `fD`:

```
julia> fD = rand(4)
4-element Array{Float64,1}:
 0.36243
 0.0790668
 0.232033
 0.975159


julia> X = unsafe_wrap(Array{Float64}, pointer(fD), 5)
5-element Array{Float64,1}:
 0.36243
 0.0790668
 0.232033
 0.975159
 9.88131e-324

julia> fD[:] = 1
1

julia> X
5-element Array{Float64,1}:
 1.0
 1.0
 1.0
 1.0
 9.88131e-324
```

(they _did_ warn us it was unsafe)

src/drstore.jl
```diff
@@ -0,0 +1,74 @@
...
+    return DRStore(tuple(D...), fD)
+end
+
+DRStore! = DRStore
+
+function Main.copy(ds::DRStore{T}) where T
```
sglyon: This should be `Base.copy`

src/drstore.jl
```diff
@@ -0,0 +1,74 @@
...
+    dims = [length(e) for e in ds.data]
+    fD = deepcopy(ds.flat)
+    ds1 = DRStore!(fD,dims)
+    return ds1
+end
+
+Main.copy!(ds0::DRStore{T}, ds1::DRStore{T}) where T = copy!(ds0.flat, ds1.flat)
```
sglyon: `Base.copy!`

src/drstore.jl
```diff
@@ -0,0 +1,74 @@
...
+
+
+function Main.clamp!(ds::DRStore{T}, ds_lb::DRStore{T}, ds_ub::DRStore{T}) where T
```
sglyon: `Base.clamp!`

src/algos/time_iteration_direct.jl
```diff
@@ -68,45 +74,49 @@ function time_iteration_direct(model, dprocess::AbstractDiscretizedProcess,
...
+
 
     while it<maxit && err>tol_η
 
         it += 1
 
         tic()
 
-        E_f = [zeros(Point{n_h},N) for i=1:number_of_smooth_drs(dprocess)]
+        set_values!(dr, [ds0.data...])
```
sglyon: Is there any kind of penalty from the `[ds0.data...]` that appears throughout? (e.g. performance, memory, or type inference penalty)

This might be pushing things too far (and please tell me if it is), but would it make sense to have `drs.data` be `NTuple{N,Vector{T}}` so we know how many elements are in it and can use `@generated` functions to unroll these `[drs.data...]` calls?

albop: For now I don't think we have any penalty. The elements of `[ds0.data...]` should be correctly inferred. `set_values!` itself  could be improved to take a tuple as argument. I doubt unrolling would provide much benefit, as the amount of code within the loop is quite big.

src/drstore.jl
albop: I think problem is not whether it is safe:
- we control object construction, and the object is immutable so user cannot alter its behaviour. Of course this assumes we add consistency checks.
- all arrays are linked to the same object, so one "view" cannot be destroyed without the other being destroyed.  By construction, the memory we access should be freed for all arrays/subarrays at the same time.
One potential problem is actually performance. From my reptilian understanding of the discussion here (https://github.com/JuliaLang/julia/issues/22849), the compiler can optimize further if we guarantee non-aliasing. Se we might slow down things a little bit with this approach (but not more than our current use of reinterpret).
In principle, SubArray should work, and we should be able to pass them to all routines (including interpolation routines). Right now it doesn't:
- our interpolation routine assume that input are contiguous arrays (i.e. take Array as arguments). We rely on this to reinterpret arrays in both directions.
- we cannot just change all arguments to AbstractArray. That would be inapropriate, because it doesn't give us any guarantee that we can reinterpret them. 
- SubArray have a flag which tells whether they are contiguous, but reinterpret doesn't work on them. 
Intuitively, the condition for us to be able to reinterpret an array, should not for it to be contiguous but for all consecutive chunks to be a multiple of the size of all datatype we wish to reinterpret it into.  I think this is the direction, which is followed in Julia 0.7, where reinterpret will return a special type of Array.  It would be natural that some SubArrays are reinterpretable too, but I don't know whether this is the case yet.

Right now, I don't totally understand how the new reintepret will work in 0.7 and it certainly won't be backward compatible.
On the other hand, the current behaviour is arguably stupid, but certainly forward compatible, unless they remove pointer, which I'm sure they won't. Maybe there is a way to actually tell the compiler we are not aliasing anything. Overall there are less factors for us to take into account with that simple solution so I'm quite inclined to stay with it for now.

src/drstore.jl
albop: I don't see anything wrong with the value `9.88131e-324`. I quite like it actually.
(joke aside: sure we need to check dimensions).

src/drstore.jl
albop: Sure. I followed the error message without thinking...

albop:
Thanks for the comments. So, this seems like the right direction. I'll make two changes:
- rename `drs.data` to `drs.grouped` so that the object is more neutral as to which is its natural layout.
- remove the indexing functions
Then I'll wait to see whether I'm hundred convinced to derive from AbstractArray{T}.
Any idea for a good name of a function to access `drs.grouped[i]` ? `συβ(drs,i)` ?

sglyon:
What if we did `drs[i, :]` for `drs.grouped[i]` (or the colon in the first position -- whichever feels more natural to you)

albop:
That would be a bit like a non square 2d array. Sounds very good to me !

albop:
As for orientation, dolo notations are (i,n) where i is exogenous index and
n endogenous one so your orientation makes sense to me. It's not classical
Julia ordering but I can live with that.

sglyon:
Ok sounds great!

albop:
Quick question : what is a nice simple way to do  `a[:] += b[:]` without allocating ?

sglyon:
I believe that `a .+= b` would work

You can always use the trusty `AXPY` Blas routine:

```julia
julia> a = [1, 2, 3.0]
b =3-element Array{Float64,1}:
  1.0
 2.0
 3.0

julia> b = [10, 10, 10.0]
3-element Array{Float64,1}:
 10.0
 10.0
 10.0

julia> Base.axpy!(1, b, a)
3-element Array{Float64,1}:
 11.0
 12.0
 13.0

julia> a
3-element Array{Float64,1}:
 11.0
 12.0
 13.0
```

albop:
Are you sure about the former ? Any ref ?

sglyon:
`a .+= b` is syntactic sugar for `broadcast!(+, a, a, b)` 

```
julia> expand(:(a .+= b))
:((Base.broadcast!)(+, a, a, b))
```

albop:
Getting back to it, not understanding much. Looks like my former self was a better version of me.

albop:
This is now redundant, as we are using (hopefully fast) views.

PR #138, opened 2018-06-06
REF: perturbate->perturb and Dolo.jl/export += tabulate

albop:
Not much here but: 
- `perturbate` becomes `perturb` as somebody told me `perturbate` means something different in english. @sglyon : can you confirm it's better like that ?
- export `tabulate`
- all results object have a `res.dr` field instead of `res.solution`, like the other result objects. Thanks @vandalm for pointing this out.

### DIFF
--- a/docs/src/algos.md
+++ b/docs/src/algos.md
@@ -144,5 +144,5 @@ find_deterministic_equilibrium
 ```
 
 ```@docs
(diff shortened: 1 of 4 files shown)


sglyon:
Sounds good to me!

Thanks for submitting.

PR #142, opened 2018-09-20
Albop/compat07

albop:
This one passes all 0.7 test on my machine, with lots of warnings to be fixed before 1.0. To be noted:
- I disabled calls to BasisMatrices (was not 1.0 compliant) and Requests (outdated). Both can be fixed in this PR
- some quickfixes for improved_time_iteration have a huge impact on performance. This will require some thinking about how to deal with reinterpretedarrays, which I propose to do independently


sglyon:
We should switch to using HTTP instead of Requests

albop:
Yes, doing that right now.

Le jeu. 20 sept. 2018 à 14:34, Spencer Lyon <notifications@github.com> a
écrit :

albop:
Incredible! World might not be quite at peace yet, but we are 1.0 compatible !

albop:
@sglyon : now that basismatrices is 1.0, we are just waiting for AxisArrays to do a fix release, then we can release Dolang.jl and Dolo.jl in turn

sglyon:
@albop great work!

PR #170, opened 2021-11-22
Albop/new iti

albop:
This is a rewrite of the time_iteration/improved_time_iteration functions.


albop:
This is a rewrite of the time_iteration and improved_time_iteration algorithms.

PR #171, opened 2021-11-25
Gqueran/tremblinghand

gabriellequeran:


albop:
Thank you @gabriellequeran ! I really like this new implementation. 
There are three more steps before I merge it:
- `my_trembling_hand_3` should be renamed to `trembling_hand!` and the former implementation removed
- imho, there are too many small functions, which complicates a bit the reading. I would keep only trembling_hand and `outer` (actually, maybe we should submit the latter to StaticArrays.jl)
- functions are missing some documentation

PR #174, opened 2021-12-09
Gqueran/work on a structure g

gabriellequeran:
The value of the partial derivative of G by x in dev.jl is not correct and must be corrected. ergodic.jl is also completed by two functions : new_transition and new_distribution.


albop:
@gabriellequeran : I've changed the PR so as to merge it into a new branch albop/new_G because we have a bit of additional work before merging in main.

PR #177, opened 2022-01-03
Albop/new g

albop:


albop:
New G object for distribution transitions.

PR #178, opened 2022-01-06
testing the function outer of ergodic.jl

gabriellequeran:


albop:
Thank you @gabriellequeran . The second test looks a bit strange. Can you simplify it a bit so as to remove the generator and the tuple construction ?

PR #179, opened 2022-01-06
Gqueran/outer

gabriellequeran:
Adding another test but this time for outer2 and commenting the function.


albop:
So here the outer2 function is tested only for 1d, right. Can you add 2d ?

albop:
OK, since you are adressing the outer2 function in another branch, I"ll merge that one as it is.

